Issue #7, opened 2022-02-25
Implement cltr_database collector


brendanheywood:
This will be a simple proof of concept collector and will have some use in making sure everything is working

* [ ] It will just dump all metrics into a new generic metric, key value table
* [ ] Admin setting for deleting old metrics after a certain time or after a certain number of records
* [ ] will have a high level report showing all the metrics which have been collected
* [ ] Because metrics may not have the same times use a line chart so we ca overlay multiple metrics and have them still work.

https://docs.moodle.org/dev/Charts_API

jaypha:
It doesn't look like it will be possible to put multiple metrics on the same chart. According to the chart API docs, all measurements must use the same 'labels', and have the same number of values as labels. In other words, all measurements on the chart must have the same timestamps.

PR #14, opened 2022-03-02
Issue #6: Add basic support for collectors.

### LINKED ISSUE #6: sub plugin type for collectors
These are the destinations for metrics to go to. Ideas for names, destination, collector, writer, exporter, observer. Out of all of these 'collector' I think makes the most sense in the wider industry and seems to clash least with existing concepts in core which already has write and exporter and observer. Again I want to make sure this prefix will never clash with other plugin types, eg coll or colectr or maybe just collect

* collect_db
* collect_awscloudwatch
* collect_azuremonitor
* collect_googlecloudmonitoring
* collect_prometheus

jaypha:
Resolves #6

Add collector_base
Add stub for cwcltr_database
Add setting to select collector
Add simple test.

Settled on 'cwcltr' as the prefix for collector subplugins.
Settings page is in 'Reports' section of Site Admin.
The database subplugin is just a stub. It will be defined for issue #7

### DIFF
--- a/classes/collector_base.php
+++ b/classes/collector_base.php
@@ -1,0 +1,70 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_cloudwatch;
+
+/**
+ * Base class for collectors.
+ *
+ * @package   tool_cloudwatch
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+abstract class collector_base {
+    public static function get_collector_classes() {
+        return \core_component::get_plugin_list_with_class('cwcltr', 'collector');
+    }
+
+    /**
+     * Returns a list of available collectors for the settings.
+     *
+     * @return array
+     */
+    public static function get_collectors_for_settings(): array {
+        $choices = [];
+        $classes = self::get_collector_classes();
+        foreach ($classes as $class) {
+            $choices[$class] = $class::get_label();
+        }
+        return $choices;
+    }
+
+    const DEFAULT_COLLECTOR = '\cwcltr_database\collector';
+    public static function get_instance(): collector_base {
+        $class = get_config('tool_cloudwatch', 'setting:destination');
+        if ($class === false) {
+            $class = self::DEFAULT_COLLECTOR;
+        }
+        return new $class();
+    }
+
+    /**
+     * Records a single value of a single metric.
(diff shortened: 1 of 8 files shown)


classes/collector_base.php
```diff
@@ -0,0 +1,70 @@
...
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_cloudwatch;
```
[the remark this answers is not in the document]
brendanheywood: across the board, only the collector is tied to cloudwatch directly

classes/collector_base.php
```diff
@@ -0,0 +1,70 @@
...
+     * @param mixed $value The metric value
+     */
+    abstract public function record_metric(string $name, int $time, $value);
+
+    /**
+     * Returns the label describing the collector.
+     * @return string
+     */
+    abstract public static function get_label(): string;
+}
```
brendanheywood, line 70: I think we will want an is_ready method, eg for cloud watch it will check if the local_aws is installed and it is configured. Similar to the is_ready for cache stores

settings.php
```diff
@@ -32,6 +32,20 @@
         'tool_cloudwatch',
         get_string('pluginname', 'tool_cloudwatch')
     );
+
     $ADMIN->add('tool_cloudwatch_reports', $settings);
+
+    if ($ADMIN->fulltree) {
+
+        $settings->add(
+            new admin_setting_configselect(
```
brendanheywood, line 41: I'd like something more like the filters or auth admin settings, ie a table where each plugin can be toggled on and off independently

brendanheywood: use the auth plugins as the example it is better as its an admin_setting not its own admin page

classes/collector_base.php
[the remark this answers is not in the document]
jaypha: done

classes/collector_base.php
```diff
@@ -0,0 +1,70 @@
...
+        $classes = self::get_collector_classes();
+        foreach ($classes as $class) {
+            $choices[$class] = $class::get_label();
+        }
+        return $choices;
+    }
+
+    const DEFAULT_COLLECTOR = '\cwcltr_database\collector';
+    public static function get_instance(): collector_base {
+        $class = get_config('tool_cloudwatch', 'setting:destination');
```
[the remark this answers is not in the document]
jaypha: I've got one modelled on dataformats and auth management

classes/plugininfo/cwcltr.php
```diff
@@ -0,0 +1,31 @@
...
+ * Subplugin class for collectors
+ *
+ * @package   tool_cloudwatch
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+
+class cwcltr extends \core\plugininfo\base {
```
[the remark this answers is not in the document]
jaypha: cltr it is.

settings.php
jaypha: I have redone it this way.

settings.php
```diff
@@ -26,12 +26,18 @@
 defined('MOODLE_INTERNAL') || die();
 
 if ($hassiteconfig) {
-    $ADMIN->add('reports', new admin_category('tool_cloudwatch_reports', 'Cloudwatch'));
+    $ADMIN->add('reports', new admin_category('tool_cloudmetrics_reports', get_string('pluginname', 'tool_cloudmetrics')));
```
brendanheywood, line 29: the settings should be nested under Admin > Plugins > Admin tool > ...

The reports themselves might later be under the reports tab though

Issue #2, opened 2022-02-24
Add metric for active users mtr_activeusers


jaypha:
Add a metric to record active users and show on an internal report.

brendanheywood:
This one should leverage the existing data from the registration data: (see pr about active users vs online users)

https://github.com/moodle/moodle/blob/master/lib/classes/hub/registration.php#L157

brendanheywood:
@jaypha one thing I'd like you to think about soonish is the relationship between metric plugins and metrics, should they be 1-1 or should they be one to many? I'm leaning towards 1-N and this one might be a good example to illustrate it. For example we could call the plugin metric_registration and it exposes all of the numbers that the hub figures out eg:

Moodle release (4.0beta+)
Number of subjects (6)
Number of users (12517)
Number of recently active users (2)
Number of role assignments (10110)
Number of posts (5039)
Number of questions (0)
Number of resources (26)

This particular metric is fairly low on the priority list as its a very slow metric rather than a near real time one but others which are more real-time might be in the same boat but still be a family of related numbers. Maybe a better example could be something around live exams, and the number of students in a quiz, or waiting to start, or who have finished and they are all pumped out of the same complicated sql query or library call.

The other thing I'd like you to think about is the relationship between metrics and the metric plugin type. It would be good if these we not tightly coupled. ie it is very convenient to have a family of metrics defined by a plugin which is an mtr_xyz plugin, but it doesn't actually need to be an mtr plugin. This would enable us to have other plugins, say mod_quiz, define their own metrics internally. This isn't a must-have requirement, just something to think about and see how viable it is and steer that way. There is a lot of parallels with the Check API for inspiration.

jaypha:
Implemented as part of #27

Issue #3, opened 2022-02-25
Metric for concurrent users aka online users


brendanheywood:
Two metrics:

* [x] Concurrent logged in users who's session has been updated within a set rolling window
* [x] We need to be very clear about users vs sessions. We are talking about sessions
* [x] use the existing definition of the online users block

PR #24, opened 2022-03-17
Issue #2: Add active users metric

jaypha:
Add logic to record active users
Add setting to control time-to-show-users.

### DIFF
--- a/metric/activeusers/classes/metric.php
+++ b/metric/activeusers/classes/metric.php
@@ -18,10 +18,12 @@ namespace metric_activeusers;
 
 use tool_cloudmetrics\metric_item;
 
+use block_online_users\fetcher;
+
 /**
  * Metric class for active users.
  *
- * @package    metric_foobar
+ * @package    metric_activeusers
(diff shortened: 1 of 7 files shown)


metric/activeusers/classes/metric.php
```diff
@@ -18,10 +18,12 @@
 
 use tool_cloudmetrics\metric_item;
 
+use block_online_users\fetcher;
```
brendanheywood, line 21: copy pasta?

brendanheywood: ah, I originally thought this was a typo but it's not. The reason I was confused is a language subtly, there is the concept of 'active users' which is the long term trend of how many users are active on the site in the last 30 days which you can pull from here:

https://github.com/moodle/moodle/blob/master/lib/classes/hub/registration.php#L157

Then there is the metric for 'online users' which is what this patch. So this patch is totally correct, and also a good priority, but it's actually issue #3 not issue #2. 

So keep working on this but rename it and realign it (might be easier as new pr)

jaypha:
Closing as incompatible with issue #27

PR #29, opened 2022-03-28
Issue #15: Task for collecting metrics.

### LINKED ISSUE #15: Decide on architecture for scheduling and time slicing for metrics
We need to decide fairly early on how each metric can declare it's granularity. ie some metrics we want daily, some we want hourly, some every fine minutes and some every minute. Behind the scenes this will be driven by cron so can't easily be run more than a minute without some hoop jumping.

So some options on the table:
* A: each metric sub plugin type manages its own scheduling and it calls a central manager with some results which then publishes to each collector. 
  * Pros: schedule is defined as a normal cron spec in the same place as all the other tasks defined in code
  * Cons: it's awkward for a plugin which might define multiple metrics with different schedules, it just moves the problem to some extent
* B: each metric defines a schedule somehow, and we have a single central task which calls into metric when it is due
  * Pros: Potentially easier config for the schedule
  * Cons: more code

I've been toggling between both but increasingly leaning towards B. So the next question then would be how do we define a schedule?

1. we could do it as a cron spec, which means out cron task needs to steal some functionality from cores cron manager. Not sure how clean this will be
2. we could say something simple like each plugin has an frequency, eg 1 minute, 60 minutes, and if it is older than that then we rerun it. Down side of this is we should aim to have metrics run on nice clean period boundaries, on the 0th minute each hour so we can better compare metrics between sites and between metrics running on the same schedule on the same site
3. I'm a growing fan of declaring a schedule via a free text string parsable by strtotime which has a large amount of expressive freedom and is very intuitive to write. This works really great for day level granulatty but not so well for minute or hourly
4. we just have a dropdown of 1 mine, 5 mins, 15 mins, hourly, 3 hours, 12 hours, 24 hours, weekly, etc and that's all you get, but they are aligned with period boundaries unlike option 2

jaypha:
Resolves #15

Add collect_metrics_task class.
Add tool_cloudmetrics_collect_metrics_test unit test class.


classes/task/collect_metrics_task.php
```diff
@@ -0,0 +1,136 @@
...
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_cloudmetrics\task;
+
+use tool_cloudmetrics\metric;
+use tool_cloudmetrics\collector;
+
+/**
+ * Controls the running of metrics within a single task. Derive from scheduled_task to enable running from
+ * the moodle task system, but it is  intendend to be run separately.
```
brendanheywood, line 24: don't quite understand this comment?

classes/task/collect_metrics_task.php
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * Perform the task
+     *
+     * @throws \Exception
+     */
+    public function execute() {
+        // This algorithm is performance important. Any opportunity to optimize should be welcome.
+
+        $tz = \core_date::get_server_timezone_object();
```
brendanheywood, line 109: This chunk of 10-15 lines I'd like a couple comments explaining the what and why of it, tz and time maths is never trivial

cli/collect_metrics.php
```diff
@@ -0,0 +1,63 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_cloudmetrics;
+
+use tool_cloudmetrics\task\collect_metrics_task;
+
+/*
+ * A script to be able to execute the task runner independantly of the task system.
```
brendanheywood, line 22: I'm not opposed to this script existing but is there any main benefit to it when we can run this like:

php admin/cli/scheduled_task.php --execute='\tool_cloudmetrics\task\collect_metrics_task'

classes/task/collect_metrics_task.php
jaypha: The comment has been modified. The original referenced something you said one, about possibly having a separate cron job to run this.

classes/task/collect_metrics_task.php
jaypha: I added a lot of comments to explain the algorithm used here.

cli/collect_metrics.php
jaypha: Removed.

classes/collector/manager.php
```diff
@@ -40,14 +40,25 @@ public static function get_collector_classes(): array {
     /**
      * Sends a metric to all enabled collectors to be recorded.
      *
-     * @param metric_item $metric_item
+     * @param metric_item $item
      */
-    public static function send_metric(metric_item $metricitem) {
+    public static function send_metric(metric_item $item) {
```
Peterburnett, line 45: Do you need to do this often enough for this to exist? this feels like a bit of an antipattern

classes/metric/builtin.php
```diff
@@ -47,7 +47,7 @@ public function get_label(): string {
      * @return int
      */
     public function get_frequency(): int {
-        return (int) get_config('tool_cloudmetrics', 'newusers_frequency');
+        return (int) get_config('tool_cloudmetrics', $this->get_name() . '_frequency');
```
Peterburnett, line 50: This feels like it should be in the base class, so you dont need to override it again in non-builtins

classes/metric/manager.php
```diff
@@ -73,17 +70,16 @@ public static function get_frequency_labels(): array {
...
     }
 
     /**
-     * Gets all metrics installed on the system.Returns an associative array of all metrics installed.
+     * Gets all metrics installed on the system. Returns an associative array of all metrics installed.
      *
      * @return array An associative array of name => metric.
      */
-    public static function get_metrics(): array {
+    public static function get_metrics(bool $enabledonly = false): array {
```
Peterburnett, line 82: Why is this false by default? I would generally assume most could would only want to care about what is currently enabled.

classes/plugininfo/cltr.php
```diff
@@ -75,9 +75,9 @@ public function set_enabled(bool $enable) {
     /**
      * Get the collector class for this plugin.
      *
-     * @return collector_base
+     * @return base
      */
-    public function get_collector(): collector_base {
+    public function get_collector(): base {
```
Peterburnett, line 80: Would it be worth having renaming base class to 'collector' and just having base definitions? that way all the things that deal with collectors in aggregate can use `collector` as their param and return types, rather than `base` which feels like a misnomer that could be a little clearer. Perhaps the same for the metric class

classes/collector/manager.php
jaypha: It was a pre-existing function, but I suppose at this point it can be removed.

classes/metric/builtin.php
jaypha: I don't want to make any assumptions about how a plugin may handle the internals of it's metrics. Nor do I want to enforce anything onto them other than the interface itself.

classes/task/collect_metrics_task.php
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * @package   tool_cloudmetrics
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+class collect_metrics_task extends \core\task\scheduled_task {
+
+    const FIVEMINUTES = 5;
```
Peterburnett, line 45: this feels like it could all be inside of the function block perhaps... not fussed either way just a thought

classes/task/collect_metrics_task.php
```diff
@@ -0,0 +1,163 @@
...
+     * Each interval (except monthly) is a whole number of the next smallest interval.
+     * So if (for example), it is determined that all 3 hour metrics are to be measured, all
+     * metrics for smaller intervals will also be measured.
+     *
+     * @param int $timediff The number of minutes since the reference time.
+     * @return int
+     */
+    public function get_frequency_cutoff(int $timediff): int {
+        if ($timediff % self::ONEWEEK == 0) {
+            return metric\manager::FREQ_WEEK;
```
Peterburnett, line 98: I think this would read a little cleaner as a switch. Less brackets control statements. idk, try it out and see what it reads like

classes/task/collect_metrics_task.php
```diff
@@ -0,0 +1,163 @@
...
+
+        // Get the time reference. The time string is in the server's timezone.
+        // It is then converted to a timestamp, which is UTC.
+        $reftime = (new \DateTime('midnight last Sunday', $tz))->getTimestamp();
+
+        // We don't need to use $tz here because timestamps are always UTC.
+        $time = $this->time ?? time();
+
+        // Get the number of minutes (rounded) that have passed since the ref time.
+        $timediff = (int) round(($time - $reftime) / 60);
```
Peterburnett, line 137: Can we replace all magic numbers here MINSECS or similar? makes it a little easier to read

classes/metric/manager.php
jaypha: It started off that way, so default behaviour is unchanged.

tests/tool_cloudmetrics_collect_metrics_test.php
```diff
@@ -0,0 +1,170 @@
...
+        $task->execute();
+    }
+
+    /**
+     * Provider function for test_execute.
+     *
+     * @return array[] Each element contains a time string, a list of frequency settings,
+     *                 and a list of metrics that should be measured.
+     */
+    public function execute_provider(): array {
```
Peterburnett, line 141: Kevin recently did some tests like this, using a generator instead, and got really wide coverage. Could we look at doing something like that?

https://github.com/catalyst/moodle-tool_mfa/blob/master/factor/token/tests/factor_test.php#L260

classes/plugininfo/cltr.php
jaypha: This naming was accepted in a previous PR, and is not part of this PR. This change is merely fixing a remnant that was not fixed where it should have been. If you think the names should be different then it needs to go in a separate issue.

classes/task/collect_metrics_task.php
jaypha: It needs to be done in a specific order. So unless PHP guarantees the order of comparisons, it needs to be done with if ... else if.

tests/tool_cloudmetrics_collect_metrics_test.php
jaypha: I don't think that would work here. How would you determine what metrics are expected to be included?

classes/task/collect_metrics_task.php
jaypha: I generally do it this way.

classes/task/collect_metrics_task.php
jaypha: Done

classes/metric/builtin.php
Peterburnett: This can just be overridden though. Anyone who wants a different frequency definition can, but it saves copy pasting when 99% of cases will remain the same

classes/metric/manager.php
Peterburnett: this is pre-prod code, what breaks if its toggled?

classes/task/collect_metrics_task.php
Peterburnett: Right, I see now, as divisions of 1 day will also be divisions of 12 hours. I think its worth commenting this, and also maybe extracting to a new function. My gut still tells me there is a more clear way, building UP from the smallest increment perhaps. especially because if you cant divide 5 minutes in equally, you can never divide 1 hour in equally. 

Doesnt look like this code is performance critical, so might not be worth the effort, but an  explanation comment for above would be great

tests/tool_cloudmetrics_collect_metrics_test.php
Peterburnett: Was just a thought, there are intricacies here that go over my head atm

Peterburnett: the generator method would just need to be smart, but its probably not worth

classes/collector/manager.php
jaypha: Done

classes/metric/builtin.php
jaypha: Done

classes/metric/manager.php
jaypha: Done

classes/task/collect_metrics_task.php
jaypha: I have reversed the test order. It is already in a function of it's own, and it describes the matryoshka nature of the intervals in the function docs.

Issue #23, opened 2022-03-17
When cleaning up old metrics in the database collector round it to the nearest day so we don't get half days of data left over


brendanheywood:

jaypha:
Would it be OK, to always move the cutoff further back in time to previous midnight, so that we always have at least the time given in the expiry value (i.e. if we have an expiry of 30 days, then we will always have at least 30 days of data.)?

It also make it easier to test.

brendanheywood:
+1

PR #30, opened 2022-03-31
Issue #23: Modify database collector cleanup task to round cutoff to …

jaypha:
…the nearest day.

Resovles #23

### DIFF
--- a/collector/database/classes/task/metrics_cleanup_task.php
+++ b/collector/database/classes/task/metrics_cleanup_task.php
@@ -40,13 +40,11 @@ class metrics_cleanup_task extends \core\task\scheduled_task {
         // We want to use the server's timezone when determining 'midnight'.
         $tz = \core_date::get_server_timezone_object();
 
-        // Adding 12 hours ensures that we round to the nearest midnight.
-        $secondsago = lib::get_metric_expiry() + (HOURSECS * 12);
+        $datestr = '-' . lib::get_metric_expiry() . ' seconds';
 
-        // Using 'midnight' sets the time to 00:00:00.
-        $datestr = '-' . $secondsago . ' seconds midnight';
-
-        $cutoff = date_create_immutable($datestr, $tz)->getTimestamp();
+        // We desire the cutoff to be at midnight, we always move further backwards in time, so we
+        // always have at least the expiry value in time's worth of data.
+        $cutoff = lib::get_midnight_of($datestr, $tz)->getTimestamp();
 
         // Purge the metrics older than this time.
         $DB->delete_records_select(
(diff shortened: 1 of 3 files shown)


collector/database/classes/task/metrics_cleanup_task.php
```diff
@@ -36,7 +37,16 @@ public function get_name() {
     public function execute() {
         global $DB;
 
-        $cutoff = time() - lib::get_metric_expiry();
+        // We want to use the server's timezone when determining 'midnight'.
+        $tz = \core_date::get_server_timezone_object();
+
+        // Adding 12 hours ensures that we round to the nearest midnight.
+        $secondsago = lib::get_metric_expiry() + (HOURSECS * 12);
```
Peterburnett, line 44: Does it actually round to nearest midnight? what if the expiry was already 1am? then it will round to 12PM

Note, I dont know how expiries are set under the hood

collector/database/tests/cltr_database_test.php
```diff
@@ -81,8 +81,14 @@ public function test_collector() {
     public function test_expiry() {
         global $DB;
 
-        $timeago = time() - (int)(19.5 * DAYSECS); // 19.5 days ago.
-        $stub = $this->get_metric_stub([1, 2, 3], $timeago, DAYSECS);
+        $tz = \core_date::get_server_timezone_object();
+
+        // Represents 10am, 20 days ago.
+        $datestr = '-20 days midnight +10 hours';
```
Peterburnett, line 87: Please also test 10PM, which will find its nearest midnight at 1200 not 0000

collector/database/classes/task/metrics_cleanup_task.php
jaypha: The expiry is set via settings. It is in seconds.
According to the PHP docs, 'midnight' sets the time component to '00:00:00', i.e. midnight.
So if the expiry was 2020-02-02 01:00:00, it will get changed to 2020-02-02 13:00:00, and then changed to
2020-02-02 00:00:00.

jaypha: https://www.php.net/manual/en/datetime.formats.relative.php

jaypha: After checking, I discovered that I misinterpreted the docs. I have made adjustments, including extra tests, to ensure that we are indeed rounding to midnight.

collector/database/tests/cltr_database_test.php
jaypha: Done

collector/database/classes/task/metrics_cleanup_task.php
jaypha: Also, it has been changed, so that it no longer rounds to the nearest midnight. It always rounds to the previous midnight.

PR #33, opened 2022-04-04
Issue #7: Add chart display

jaypha:
Add chart display page
Add listing to site administration reports

This PR includes a CLI script to send fake values to the collectors.

```
php admin/tool/cloudmetrics/cli/add_metrics.php --metric=activeusers
```

To run it you need to have a CFG value set to a non empty value. This prevents accidental running in prod environments.
The setting is `$CFG->config_php_settings['tool_cloudmetrics_allow_add_metrics'] =1;`.

You run it with a name to 'fake' the metric. i.e. `add_metrics.php --name=activeusers`.
There are other options. Use `--help` to list them.

After this you should be able to see the values on the charts page. (Site Administration -> Reports -> Reports -> Cloudmetrics charts).

IMO, the Charts API is very primitive, and recommend that, at some future date, something like D3 be used instead.

### DIFF
--- a/cli/add_metrics.php
+++ b/cli/add_metrics.php
@@ -1,0 +1,77 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+use tool_cloudmetrics\metric\test_metric;
+use tool_cloudmetrics\collector;
+
+/**
+ * Generate mock metric data to send to collectors.
+ *
+ * @package   tool_cloudmetrics
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+define('CLI_SCRIPT', true);
+
+require_once(__DIR__ . '/../../../../config.php');
+require_once($CFG->libdir.'/clilib.php');
+
+// Get cli options.
+list($options, $unrecognized) = cli_get_params(
+    [
+        'help' => false,
+        'name' => false,
+    ], [
+        'h' => 'help',
+        'n' => 'name',
+    ]
+);
+
+if ($unrecognized) {
+    $unrecognized = implode("\n  ", $unrecognized);
+    cli_error(get_string('cliunknowoption', 'admin', $unrecognized));
+}
+
+if ($options['help']) {
+    $help =
+        "Add metrics.
+
+Generate mock metric data to fill collectors with.
+Do not use with production environments.
+
+Options:
+-h, --help               Print out this help
+-n, --name               The name of the metric being mocked.
+
+Example:
+\$sudo -u www-data /usr/bin/php admin/tool/cloudmetrics/cli/add_metrics.php --name=activeusers
+";
+
+    echo $help;
+    die;
+}
+
+
+
+$metric = new test_metric();
+$metric->name = $options['name'];
+
+for ($i = 0; $i < 100; ++$i) {
+    $item = $metric->get_metric_item();
+    collector\manager::send_metric($item);
+}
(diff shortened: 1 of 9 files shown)


cli/add_metrics.php
```diff
@@ -0,0 +1,77 @@
...
+    $unrecognized = implode("\n  ", $unrecognized);
+    cli_error(get_string('cliunknowoption', 'admin', $unrecognized));
+}
+
+if ($options['help']) {
+    $help =
+        "Add metrics.
+
+Generate mock metric data to fill collectors with.
+Do not use with production environments.
```
Peterburnett, line 55: is it worth adding some checks here to prevent prod runs? We did this with datawash 5 different ways (really) to be safe, but it can be just 1 perhaps. Might not be worth.

There *may* be a security influence here, in that someone may be able to inject naughty data with a system compromise, and that may be used to divert monitoring attention to the fake load spike or similar, while the attacker targets another system.

collector/database/chart.php
```diff
@@ -0,0 +1,76 @@
...
+ * Shows a chart of recorded metrics.
+ *
+ * @package   cltr_database
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+require_once(__DIR__.'/../../../../../config.php');
+require_once($CFG->libdir.'/adminlib.php');
```
Peterburnett, line 29: I think CI wants spaces before/after dots. Only a warning but worth mentioning

collector/database/chart.php
```diff
@@ -0,0 +1,76 @@
...
+
+admin_externalpage_setup('cltr_database_chart');
+
+$context = context_system::instance();
+
+$url = new moodle_url('/admin/tool/cloudmetrics/collector/database/chart.php');
+
+$PAGE->set_context($context);
```
Peterburnett, line 37: I think admin_externalpage_setup gives you the context for free

collector/database/chart.php
```diff
@@ -0,0 +1,76 @@
...
+
+$select = new \single_select(
+    $url,
+    'metric',
+    $metriclabels,
+    $metricname
+);
+$select->set_label(get_string('select_metric_for_display', 'cltr_database'));
+
+$records = $DB->get_records('cltr_database_metrics', ['name' => $metricname], 'time asc', 'id, time, value');
```
Peterburnett, line 56: Just a thought, this feels like something that should come in via the metric class itself, eg $metric->get_records()

This means you will only load metrics for the classes you can load, so if you uninstall a subplugin, its data would no longer be exposed. But it also feels more right to get metric data via the metric as part of its public API

collector/database/chart.php
```diff
@@ -0,0 +1,76 @@
...
+    $labels[] = userdate($record->time, get_string('strftimedatetimeshort'), $CFG->timezone);
+}
+
+$chartseries = new \core\chart_series($metriclabels[$metricname], $values);
+
+$chart = new \core\chart_line();
+$chart->add_series($chartseries);
+$chart->set_labels($labels);
+
+$output = $PAGE->get_renderer('core');
```
Peterburnett, line 72: $OUTPUT is the core renderer by default, so you can just use the $OUTPUT global for all of these

collector/database/settings.php
```diff
@@ -28,6 +28,18 @@
 use cltr_database\output\admin_setting_days_configduration;
 
 if ($hassiteconfig) {
+    $ADMIN->add('reports', new admin_category('cltr_database_reports', get_string('pluginname', 'cltr_database')));
```
Peterburnett, line 31: What other reports should live in this? If its just the 1, it would be better to just add it directly to the 'reports' category

Peterburnett:
One more thing, new admin_settings need a version bump, otherwise they will be null in code when called and throw naughty errors

collector/database/chart.php
jaypha: The data is stored in the database collector. The metrics have no knowledge of this. The metrics only generate the data and pass it on the collectors.

collector/database/chart.php
jaypha: I had a look at the code. It sets the context to null, but nothing else.

collector/database/chart.php
Peterburnett: Then perhaps a collector API that accepts an array of metrics, and returns a multidimension array of all the data against each metric. With something so OO, it feels weird to mix in DB calls that collect data across mutliple objects

cli/add_metrics.php
jaypha: I added a simple $CFG check. You need to set a config value $CFG->config_php_settings['tool_cloudmetrics_allow_add_metrics'], to be able to run the script.
I basically copied check #4 from datacleaner.

collector/database/chart.php
jaypha: done

collector/database/chart.php
jaypha: done

collector/database/chart.php
jaypha: done

collector/database/settings.php
jaypha: done

jaypha:
Here is the latest update.
I have added some instructions to the main comment on how to generate fake data.

PR #42, opened 2022-04-08
Issue #35, 38: Settings Tweaks

### LINKED ISSUE #35: Add a link from metrics settings table to all the reports
Make a convenient link from here into the report

image

If there are multiple collectors then there will be multiple columns

Allow the link even if the metic is disabled as there might be old data in there. Make it a gray link though

### LINKED ISSUE #38: Each metric should have a method to return a blurb about exact what it is
* [x] Show this in the admin settings metric table
* [x] show it at the top of the db chart page
* [x] add this as a string for the 3 built in metrics

jaypha:
Resolves #35 
Resolves #38
Add description text to metrics.
Add description to chart page, and to settings table.
Add Link to chart page from settings table.
Collapse action links into icons, for metrics and collectors.

### DIFF
--- a/classes/admin_setting_manage_collectors.php
+++ b/classes/admin_setting_manage_collectors.php
@@ -97,7 +97,7 @@ class admin_setting_manage_collectors extends \admin_setting {
         $table = new \html_table();
         $table->head  = array($txt->name, $txt->uninstall, $txt->actions);
         $table->align = array('left', 'center', 'center', 'center', 'center');
-        $table->attributes['class'] = 'manageformattable generaltable admintable';
+        $table->attributes['class'] = 'manageformattable generaltable admintable w-auto';
         $table->data  = array();
 
         foreach ($formats as $format) {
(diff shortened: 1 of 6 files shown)


classes/admin_setting_manage_collectors.php
```diff
@@ -88,26 +88,26 @@ public function is_related($query) {
...
 
-        $txt = get_strings(array('settings', 'name', 'enable', 'disable', 'default'));
+        $txt = get_strings(array('settings', 'name', 'enable', 'disable', 'default', 'actions'));
         $txt->uninstall = get_string('uninstallplugin', 'core_admin');
 
         $table = new \html_table();
-        $table->head  = array($txt->name, $txt->enable, $txt->uninstall, $txt->settings);
+        $table->head  = array($txt->name, $txt->uninstall, $txt->actions);
         $table->align = array('left', 'center', 'center', 'center', 'center');
-        $table->attributes['class'] = 'manageformattable generaltable admintable w-auto';
```
brendanheywood, line 101: can you please rebase to keep the w-auto I added

jaypha: done

PR #44, opened 2022-04-11
Issue #36: Improve settings interface.

### LINKED ISSUE #36: Admin setting table tweaks
1) add in the component name of the thing which declare the metrics
2) Id like to see the schedule for each metric in this table. The easy version of this is just show it as text.
 Bonus points: The more robust way is to make this an inline editable which replaces the admin setting. https://docs.moodle.org/dev/Inplace_editable
3) each metric should declare a place to go for its settings and that should be in the table. Doesn't make heaps of sense for the build in ones but will be better later
image

jaypha:
Resolves #36 

Add plugin name to metric row.
Move frequency selector to the table as an inline editable.


brendanheywood:
This is really neat :)

PR #46, opened 2022-04-13
Issue #8: add support for AWS Cloudwatch

### LINKED ISSUE #8: Implement cltr_cloudwatch collector
* [ ] Export all metrics to cloud watch
* [ ] In the metrics admin settings link of to a deep link into aws land to where the metric can be seen
* [ ] Add setup notes to readme file.

jaypha:
Resolves #8


collector/cloudwatch/readme.md
```diff
@@ -0,0 +1,49 @@
+# cltr_cloudwatch
+
```
brendanheywood, line 2: This is small enough that I would inline this into the main plugin readme instead, we can split it later if and when we have a ton of collectors and it is too cluttered

collector/cloudwatch/settings.php
```diff
@@ -0,0 +1,88 @@
...
+        $settings->add(new admin_setting_heading('cltr_cloudwatch_settings', '',
+            get_string('pluginnamedesc', 'cltr_cloudwatch')));
+
+        // AWS settings.
+        $settings->add(new admin_setting_heading('cltr_cloudwatch_aws',
+            get_string('awssettings', 'cltr_cloudwatch'),
+            get_string('awssettings_desc', 'cltr_cloudwatch')
+        ));
+
+        $regionoptions = [
```
brendanheywood, line 40: in local_aws there is a region selection admin setting admin_settings_aws_region

collector/cloudwatch/classes/collector.php
```diff
@@ -94,55 +94,6 @@ public function is_ready(): bool {
...
-                    'MetricName' => $metric->get_name(),
-                    'Namespace' => $namespace,
-                ],
-                'Period' => 30,
-                'Stat' => 'Average',
-                'Unit' => $metric->unit ?? 'Count',
-            ],
-            'ReturnData' => true
-        ];
+        return false;
```
brendanheywood, line 97: should get_metrics be something unique to the db collection and not in the base class? I can't see a need for every querying metrics from each collector and in many cases it's probably not possible or easy

collector/cloudwatch/readme.md
jaypha: Done

collector/cloudwatch/settings.php
jaypha: Done

collector/cloudwatch/classes/collector.php
jaypha: Done

collector/cloudwatch/version.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+defined('MOODLE_INTERNAL') || die();
+
+$plugin->version = 2022041100;
+$plugin->requires = 2017051500;    // Our lowest supported Moodle (3.3.0).
+$plugin->component = 'cltr_cloudwatch';
+
+$plugin->dependencies = array(
+    'local_aws' => 2022033100
```
brendanheywood, line 33: also I'm on the fence about requiring it. Right now this is totally fine but at same stage when we add more collectors this needs to be a runtime dependency check similar to objectfs as people will want to run it in a different cloud. So lets remove this and make it a check in is_ready

brendanheywood: https://github.com/catalyst/moodle-tool_objectfs/blob/master/classes/local/store/s3/client.php#L378-L394

README.md
```diff
@@ -57,9 +57,51 @@ We plan to have a very simple internal record of metrics with a limited data ret
...
+    }
+
+Or the user can be assigned the `CloudWatchFullAccess`
+predefined policy.
+
+Ideally, the execution environent will be configured with
+the IAM credentials. However, the following config can
+also be used. Substitute the value strings with your access keys.
+
+    $CFG->cltr_cloudwatch['credentials']['key'] = '<Access key ID>';
```
brendanheywood, line 92: Can you please convert this to use standard admin settings so they can *optionally* be in the UI, but can still be forced in config.php and update the docs here to be like

```
$CFG->forced_plugin_settings['cltr_cloudwatch']['credentials'] = 'blah';
```

brendanheywood: Use the password admin setting for secrets, it does some magic when the value is forced so you can't see it in the gui

collector/cloudwatch/settings.php
```diff
@@ -0,0 +1,77 @@
...
+        $settings->add(new admin_setting_heading('cltr_cloudwatch_general',
+            get_string('generalsettings', 'cltr_cloudwatch'),
+            get_string('generalsettings_desc', 'cltr_cloudwatch')
+        ));
+
+        // Namespace.
+        $settings->add(new admin_setting_configtext('cltr_cloudwatch/namespace',
+            get_string('namespace', 'cltr_cloudwatch' ),
+            get_string('namespace_desc', 'cltr_cloudwatch'),
+            'Moodle', PARAM_TEXT));
```
brendanheywood, line 61: wondering if this should default to the current domain moodle is running on, OR even better make it optional and default it to off, and if it is empty then it defaults to the domain so that this will keep working most of the time if we ever change domains. I can imagine this being run with many clients with multiple environments in the same AWS account

collector/cloudwatch/settings.php
```diff
@@ -0,0 +1,77 @@
...
+        $settings->add(new \local_aws\admin_settings_aws_region('cltr_cloudwatch/awsregion',
+            get_string('awsregion', 'cltr_cloudwatch' ),
+            get_string('awsregion_desc', 'cltr_cloudwatch'),
+            'ap-southeast-2'
+        ));
+
+        $settings->add(new admin_setting_configselect('cltr_cloudwatch/awsversion',
+            get_string('awsversion', 'cltr_cloudwatch' ),
+            get_string('awsversion_desc', 'cltr_cloudwatch'),
+            '2010-08-01', ['2010-08-01' => '2010-08-01', 'latest' => 'latest']));
```
brendanheywood, line 49: At minimum probably need a url in the desc help about what this practically means. Given that it is locked to code I'm wondering why we wouldn't even lock it to the latest version supported by the aws sdk versionwe are depending on?

collector/cloudwatch/settings.php
```diff
@@ -0,0 +1,77 @@
...
+        ));
+
+        // Namespace.
+        $settings->add(new admin_setting_configtext('cltr_cloudwatch/namespace',
+            get_string('namespace', 'cltr_cloudwatch' ),
+            get_string('namespace_desc', 'cltr_cloudwatch'),
+            'Moodle', PARAM_TEXT));
+
+        $envoptions = [
+            'dev' => 'dev',
```
brendanheywood, line 64: lang strings?

jaypha: I wonder if a better way is to open it up. This setting is about dimensions, so maybe have it user definable as a set of <name>=<value> pairs with Environment=Dev as the default.

collector/cloudwatch/settings.php
jaypha: I copied this across from local_cloudwatch, but I think it should be in-code defined.

collector/cloudwatch/settings.php
jaypha: I'm currently setting it to the value of $CFG->wwwroot. I'm thinking that any subdirectories should be included. The 'https://' could be chopped, but it works with it included.

README.md
jaypha: Do you want me to add to add settings for these?

collector/cloudwatch/settings.php
brendanheywood: Possible characters are: alphanumeric characters (0-9A-Za-z), period (.), hyphen (-), underscore (_), forward slash (/), hash (#), and colon (:).

https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_concepts.html#Namespace

+1 to trim the http schema off and keep the directory if any

README.md
brendanheywood: Yes

brendanheywood: example 

https://github.com/catalyst/moodle-search_elastic/blob/MOODLE_310_STABLE/settings.php#L55-L61

collector/cloudwatch/settings.php
jaypha: done

collector/cloudwatch/settings.php
jaypha: done

collector/cloudwatch/settings.php
jaypha: done

collector/cloudwatch/classes/collector.php
```diff
@@ -0,0 +1,88 @@
...
+        global $CFG;
+
+        if (is_null(self::$client)) {
+            $clientconfig = [
+                'region' => get_config('cltr_cloudwatch', 'awsregion'),
+                'version' => lib::AWS_VERSION,
+            ];
+
+            // Add AWS credentials if specified in CFG. IAM role and environment would apply if they are not set.
+            if (isset($CFG->forced_plugin_settings['cltr_cloudwatch']['credentials'])) {
```
brendanheywood, line 51: forced_plugin_settings is how you force set it, but for reads this comes through as normal get_config. If you read it from forced_plugin_settings then it will only be able to be set via forced_plugin_settings and it will ignore if set in the ui

jaypha: done

collector/cloudwatch/lang/en/cltr_cloudwatch.php
```diff
@@ -0,0 +1,61 @@
...
+$string['awskey_desc'] = 'The access key ID assigned to the IAM user.';
+$string['awssecret'] = 'AWS access secret';
+$string['awssecret_desc'] = 'The secret access key assigned to the IAM user.';
+
+// Missing requirements.
+$string['unsatisfied_requirements'] = 'Unsatisfied Requirements';
+$string['aws:installneeded'] = 'Plugin \'local_aws\' of version '. \cltr_cloudwatch\lib::LOCAL_AWS_VERSION . ' is required.';
+$string['aws:upgradeneeded'] = 'Plugin \'local_aws\' needs to be upgraded to version '. \cltr_cloudwatch\lib::LOCAL_AWS_VERSION . '.';
+
+$string['Dev'] = 'Dev';
```
brendanheywood, line 58: string id's I think should be all lower

The acronyms should probably be upper case, UAT, QAT, but Dev and Prod are words and can be initial caps

PR #48, opened 2022-04-14
Issue #45: Set in-code defaults for metric frequencies.

### LINKED ISSUE #45: Bug: inline editables need to have a default metric frequency
This will die in install as well as when settings are unset:

php admin/cli/cfg.php --component=tool_cloudmetrics --name=onlineusers_frequency --unset

Coding error detected, it must be fixed by a programmer: Options for select element must contain an option for the specified value

http://moodle.local/admin/category.php?category=tool_cloudmetrics_reports

I think each metric should declare in code what its default frequency is rather than just relying on the admin settings like before

jaypha:
Resolves #45 

Rearrange metric settings to be in alphabetical order.
Fix bad label for New users setting.
Version bump

### DIFF
--- a/classes/metric/base.php
+++ b/classes/metric/base.php
@@ -59,8 +59,19 @@ abstract class base {
      * @return int
      */
     public function get_frequency(): int {
-        return (int) get_config('tool_cloudmetrics', $this->get_name() . '_frequency');
+        $freq = (int) get_config('tool_cloudmetrics', $this->get_name() . '_frequency');
+        if ($freq == 0) {
+            $freq = $this->get_frequency_default();
+        }
+        return $freq;
     }
+
+    /**
+     * Return the default setting value for the metric frequency.
+     *
+     * @return int
+     */
+    abstract public function get_frequency_default(): int;
 
     /**
      * Set frequency of the metric's sampling.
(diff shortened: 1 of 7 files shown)


classes/metric/manager.php
```diff
@@ -74,6 +74,13 @@ public static function get_frequency_labels(): array {
         ];
     }
 
+    /** @var int[] Default settings for builtin metric frequencies. */
```
keevan, line 77: I don't think the type correctly reflects the content. I don't remember the correct one specifically, but it might be something like `int[string]` instead?

keevan (commented):
Other that minor styling feedback, everything else seems fine.

classes/metric/base.php
```diff
@@ -59,9 +59,20 @@ abstract public function get_plugin_name(): string;
      * @return int
      */
     public function get_frequency(): int {
-        return (int) get_config('tool_cloudmetrics', $this->get_name() . '_frequency');
+        $freq = (int) get_config('tool_cloudmetrics', $this->get_name() . '_frequency');
+        if ($freq == 0) {
```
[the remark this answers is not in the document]
jaypha: done

classes/metric/manager.php
jaypha: done

PR #63, opened 2022-05-05
Issue #60: Wrap collector call with try .. catch.

jaypha:


classes/collector/manager.php
```diff
@@ -46,8 +46,12 @@ public static function send_metrics(array $items) {
...
         foreach ($plugins as $plugin) {
             $collector = $plugin->get_collector();
-            if ($collector->is_ready()) {
-                $collector->record_metrics($items);
+            try {
+                if ($collector->is_ready()) {
+                    $collector->record_metrics($items);
+                }
+            } catch (\Exception $e) {
+                error_log('Collector ' . $plugin->name . ' failed. "' . $e->getMessage() . '"');
```
brendanheywood: This can just be ignored with the @ directive or make it a debugging NORMAL call

Also made a followup to expose this to alerts https://github.com/catalyst/moodle-tool_cloudmetrics/issues/64

macgirvin: I believe I read in the php 8.1 release notes that @ is being deprecated - though it will take a few years to get rid of it.

Issue #64, opened 2022-05-05
Add check for failing metric pushes


brendanheywood:
https://github.com/catalyst/moodle-tool_cloudmetrics/pull/63/files

As well as emitting an error log, if a collector fails then set a config var for it with a time stamp. If later on the metrics work then unset the timestamp.

Then make a Check API which show if a collector has been failing, and if so for how long it has been down

PR #68, opened 2022-05-05
Add a metric collector status check

macgirvin:
Issue #64 store timestamp of plugin failure and create API to query failures.

  [x] Store timestamp on first error and clear on success.
  [x] Create check API interface

### DIFF
--- a/classes/check/collectorcheck.php
+++ b/classes/check/collectorcheck.php
@@ -1,0 +1,98 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Collector failure tests for Check API
+ *
+ * @package tool_cloudmetrics
+ * @author Mike Macgirvin (mikemacgirvin@catalyst-au.net)
+ * @license http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ * @copyright Catalyst IT
+ */
+
+namespace tool_cloudmetrics\check;
+
+use core\check\check;
+use core\check\result;
+use tool_cloudmetrics\collector\manager;
+use tool_cloudmetrics\plugininfo\cltr;
+
+class collectorcheck extends check {
+
+    /**
+     * Constructor
+     */
+    public function __construct() {
+        $this->id = 'checkcollectorcheck';
+        $this->name = get_string('checkcollectorcheck', 'tool_cloudmetrics');
+    }
+
+    /**
+     * A link to a place to action this
+     *
+     * @return action_link|null
+     */
+    public function get_action_link(): ?\action_link {
+        return new \action_link(
+            new \moodle_url('/admin/category.php?category=tool_cloudmetrics_reports'),
+            get_string('managelink', 'tool_cloudmetrics'));
+    }
+
+    /**
+     * Return result
+     * @return result
+     */
+    public function get_result() : result {
+        global $CFG;
+
+        $failures = false;
+        $warnings = false;
+        $messages = [];
+
+        $names = cltr::get_ready_plugin_names();
+        foreach ($names as $name) {
+            $result = get_config('tool_cloudmetrics', manager::STATUS_PREFIX . $name);
+            if (! $result) {
+                $warnings = true;
+                $messages[] = get_string('collector_never', 'tool_cloudmetrics', $name);
+                continue;
+            }
+            if ($result === 'pass') {
+                $messages[] = get_string('collector_passed', 'tool_cloudmetrics', $name);
+                continue;
+            }
+            $failures = true;
+            $summarytemplate = get_string('collector_failed', 'tool_cloudmetrics',
+                ['name' => $name, 'time' => userdate((int) $result)]);
+        }
+
+        if ($messages) {
+            $failuretype = result::OK;
+            if ($warnings) {
+                $failuretype = result::WARNING;
+            }
+            if ($failures) {
+                $failuretype = result::ERROR;
+            }
+            // This result contains the enumerated detail of each test.
+            return new result($failuretype, implode('<br>', $messages));
+        }
+
+        return new result(result::INFO, get_string('no_collectors', 'tool_cloudmetrics'), '');
+    }
+
+
+}
--- a/classes/collector/manager.php
+++ b/classes/collector/manager.php
@@ -28,6 +28,9 @@ use tool_cloudmetrics\plugininfo\cltr;
  * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 class manager {
+
+    const STATUS_PREFIX = 'collector_status:';
+
     /**
      * Get all the collector classes.
      *
@@ -49,9 +52,16 @@ class manager {
             try {
                 if ($collector->is_ready()) {
                     $collector->record_metrics($items);
+                    // Clear any failure timestamps for this plugin.
+                    set_config(self::STATUS_PREFIX . $plugin->name, 'pass', 'tool_cloudmetrics');
                 }
             } catch (\Exception $e) {
                 debugging('Collector ' . $plugin->name . ' failed. "' . $e->getMessage() . '"');
+                // Store plugin name with timestamp of initial failure to track when problem first arose.
+                $lastfail = get_config('tool_cloudmetrics_failure', $plugin->name);
+                if ($lastfail === false || $lastfail === 'pass') {
+                    set_config(self::STATUS_PREFIX . $plugin->name, time(), 'tool_cloudmetrics');
+                }
             }
         }
     }
--- a/lang/en/tool_cloudmetrics.php
+++ b/lang/en/tool_cloudmetrics.php
@@ -83,6 +83,11 @@ $string['one_week'] = '1 week';
 $string['one_fortnight'] = '1 fortnight';
 $string['one_month'] = '1 month';
 
-// Error messages.
+// Error and status messages.
 $string['metric_not_found'] = 'Metric \'{$a}\' not found';
-
+$string['collector_failed'] = 'Collector \'{$a->name}\' failed {$a->time}';
+$string['collector_passed'] = 'Collector \'{$a}\' succeeded';
+$string['collector_never'] = 'Collector \'{$a}\' has never executed';
+$string['no_collectors'] = 'No active collectors';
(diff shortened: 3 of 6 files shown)


classes/collector/manager.php
```diff
@@ -49,9 +49,15 @@ public static function send_metrics(array $items) {
             try {
                 if ($collector->is_ready()) {
                     $collector->record_metrics($items);
+                    // Clear any failure timestamps for this plugin.
+                    unset_config($plugin->name, 'tool_cloudmetrics_failure');
```
brendanheywood: Maybe it is better to store an explicit 'pass' so we can detect the different between 'unknown' vs 'good' in the check we expose

Also check if we need to set it and only set it if needed

lang/en/tool_cloudmetrics.php
```diff
@@ -85,4 +85,7 @@
 
 // Error messages.
 $string['metric_not_found'] = 'Metric \'{$a}\' not found';
-
+$string['collector_failed'] = 'Metric \'{$a}\' failed at %s';
+$string['no_issues'] = 'No issues';
+$string['checkcollectorcheck'] = 'Collector Failures';
+$string['managelink'] = 'admin/settings.php?section=tool_cloudmetrics';
```
brendanheywood: urls don't need to be translated so this doesn't need to be a string, just use a moodle_url where it is needed

classes/check/collectorcheck.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Return result
+     * @return result
+     */
+    public function get_result() : result {
+
+        $failures = false;
+        $tests = get_config('tool_cloudmetrics_failure');
```
brendanheywood: The $plugin param should be a plugin component name like 'tool_cloudmetrics' it is a very slight abuse of the config api to use it like this.

The two alternatives are:
1) you save each collector_failed value under the plugin it belongs to
2) you save it under tool_cloudmetrics but prefix each key with something like 'collector_status'. I'd prefer _status over _failed because it will also contain positives as well as negatives

classes/check/collectorcheck.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * A link to a place to action this
+     *
+     * @return action_link|null
+     */
+    public function get_action_link(): ?\action_link {
+        return new \action_link(
+            new \moodle_url('/admin/tool/cloudmetrics'),
```
brendanheywood: this link and it's labek I think are the wrong way around, it doesn't work here:

image

This link should take you to the place where you can resolve the issue, which in this case is the settings page ie /admin/category.php?category=tool_cloudmetrics_reports

lang/en/tool_cloudmetrics.php
macgirvin: I copied what was done in the example you sent me earlier:

https://github.com/catalyst/moodle-auth_saml2/blob/MOODLE_39_STABLE/classes/check/certificateexpiry.php

but I'll fix it.

classes/check/collectorcheck.php
```diff
@@ -0,0 +1,78 @@
...
+        $failures = false;
+        $tests = get_config('tool_cloudmetrics_failure');
+        if ($tests) {
+            foreach ($tests as $name => $result) {
+                if ($result === 'pass') {
+                    continue;
+                }
+                $failures = true;
+                $summarytemplate = get_string('collector_failed', 'tool_cloudmetrics', $name);
+                $summary = sprintf($summarytemplate, date('Y-m-d H:i:s', (int) $result));
```
brendanheywood: Moodle has its own date api, use userdate() instead. You can pass it a format which is localized, for use cases like this I prefer long form human readable dates which is what it defaults to.

https://docs.moodle.org/dev/Time_API#Time_API.27s_for_current_user

classes/check/collectorcheck.php
```diff
@@ -0,0 +1,78 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Collector failure tests for Check API
+ *
+ * @package admin-tool_cloudmetrics
```
brendanheywood: This should just be the component name tool_cloudmetrics (and ci should have warned about this....)

brendanheywood: To see this manually run this check:

php local/moodlecheck/cli/moodlecheck.php -p=admin/tool/cloudmetrics/

This plugin is full of warnings, I'm not sure why they are not being reported on, will chase that

classes/check/collectorcheck.php
```diff
@@ -0,0 +1,78 @@
...
+                $summarytemplate = get_string('collector_failed', 'tool_cloudmetrics', $name);
+                $summary = sprintf($summarytemplate, date('Y-m-d H:i:s', (int) $result));
+                return new result(result::ERROR, $summary, '');
+            }
+
+        }
+
+        if (!$failures) {
+            $summary = get_string('no_issues', 'tool_cloudmetrics');
+            return new result(result::OK, $summary, '');
```
brendanheywood: Each check has a summary and a details string. In cases like this it is often good to have the summary be good / bad, but the details has a list of each thing it checked and whether it is good or bad.

classes/collector/manager.php
```diff
@@ -49,9 +49,16 @@ public static function send_metrics(array $items) {
             try {
                 if ($collector->is_ready()) {
```
brendanheywood, line 53: We are missing logic here, if it is not enabled then it shouldn't be checked. If it was checked and was failing and you disable it then it is still failing

brendanheywood: I also think there might be a logic but with the original code, when sending metrics the collector must be both is_enabled and is_ready

classes/check/collectorcheck.php
brendanheywood: option 2 is better me thinks

classes/check/collectorcheck.php
```diff
@@ -0,0 +1,88 @@
...
+                if (strpos($key, manager::STATUS_PREFIX) === 0) {
+                    $name = substr($key, strlen(manager::STATUS_PREFIX));
+                    if ($result === 'pass') {
+                        $messages[] = get_string('collector_passed', 'tool_cloudmetrics', $name);
+                        continue;
+                    } else {
+                        $failures = true;
+                        $summarytemplate = get_string('collector_failed', 'tool_cloudmetrics', $name);
+                        // Fill in the time it first failed.
+                        $messages[] = sprintf($summarytemplate, userdate((int) $result));
```
brendanheywood: You don't need the sprintf,this can be one line as get_string can accept params directly eg

get_string('collector_failed', 'tool_cloudmetrics', $name, ['foo' => 'bar']);

'Metric \'{$a->foo}\' failed %s'

https://docs.moodle.org/dev/String_API#Run-time_parameters

classes/check/collectorcheck.php
```diff
@@ -0,0 +1,88 @@
...
+                }
+            }
+        }
+
+        if ($failures) {
+            // This result contains the enumerated detail of each test.
+            return new result(result::ERROR, implode('<br>', $messages));
+        }
+
+        return new result(result::OK, get_string('no_issues', 'tool_cloudmetrics'), '');
```
brendanheywood: Even if it is a success I think the details should be the same verbose info

classes/check/collectorcheck.php
```diff
@@ -0,0 +1,88 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Collector failure tests for Check API
+ *
+ * @package tool_cloudmetrics
```
brendanheywood, line 20: Please add yourself using the `@author` you deserve the credit

lang/en/tool_cloudmetrics.php
```diff
@@ -85,4 +85,8 @@
 
 // Error messages.
 $string['metric_not_found'] = 'Metric \'{$a}\' not found';
-
+$string['collector_failed'] = 'Metric \'{$a}\' failed %s';
+$string['collector_passed'] = 'Metric \'{$a}\' succeeded';
+$string['no_issues'] = 'No issues';
+$string['checkcollectorcheck'] = 'Collector Failures';
```
brendanheywood: Also this label doesn't mean much out of context if you are looking at the status report. I think it could be more like 'Cloud metric collectors'

lang/en/tool_cloudmetrics.php
```diff
@@ -85,4 +85,8 @@
 
 // Error messages.
 $string['metric_not_found'] = 'Metric \'{$a}\' not found';
-
+$string['collector_failed'] = 'Metric \'{$a}\' failed %s';
+$string['collector_passed'] = 'Metric \'{$a}\' succeeded';
+$string['no_issues'] = 'No issues';
```
brendanheywood: I think for a check this simple you can probably just put the verbose details straight into the summary. So instead of 'No issues' it would say roughly 'db:ok, cloudwatch: ok' (with formatting etc)

image

classes/check/collectorcheck.php
```diff
@@ -0,0 +1,88 @@
...
+     * @return result
+     */
+    public function get_result() : result {
+
+        $failures = false;
+        $messages = [];
+
+        $tests = get_config('tool_cloudmetrics');
+        if ($tests) {
+            foreach ($tests as $key => $result) {
```
brendanheywood: Also wouldn't it be better to start with the list of collectors and then loop through them looking in the config? Not sure the config itself should be the canonical truth here

classes/collector/manager.php
macgirvin: On line 49 there is this: 

     $plugins = cltr::get_enabled_plugins();

At the collector level we only have is_ready() available. Looks like I will need to duplicate a fair bit of this code on the check side of things so that if somebody uninstalls a plugin we'll ignore the results for any of its children. Should I also delete them from the config?

brendanheywood: ok lets ignore this for now I might retest this when this has landed and make a followup issue

classes/check/collectorcheck.php
```diff
@@ -0,0 +1,116 @@
...
+
+        return new result(result::INFO, get_string('no_collectors', 'tool_cloudmetrics'), '');
+    }
+
+    /**
+     * Return a list of currently active plugin names.
+     *
+     * @return array
+     */
+    public function get_active_metrics() {
```
brendanheywood: This is getting collectors, not metrics, I think it should be renamed to  get_*_collectors 

I don't think this should belong in the check object it should be in the same place as get_enabled_plugins.

Also it's not clear what 'active' means, internally we are checking for is_ready so a better name would be get_ready_collectors

lang/en/tool_cloudmetrics.php
```diff
@@ -85,4 +85,9 @@
 
 // Error messages.
 $string['metric_not_found'] = 'Metric \'{$a}\' not found';
-
+$string['collector_failed'] = 'Metric \'{$a->name}\' failed {$a->time}';
```
brendanheywood: These 3 strings are for collectors, not for metrics

Issue #75, opened 2022-05-11
cltr_cloudwatch and most collectors should default off


brendanheywood:
It must be configured to work.

So default off, if forced config then we can force enable it too

PR #81, opened 2022-05-11
  Issue #75; Disable AWS cloudwatch plugin for cloudmetrics by default.

macgirvin:
Issue #75; Disable AWS cloudwatch plugin for cloudmetrics by default.
  It is automatically enabled when installed with cloudmetrics.

### DIFF
--- a/classes/plugininfo/cltr.php
+++ b/classes/plugininfo/cltr.php
@@ -76,7 +76,7 @@ class cltr extends \core\plugininfo\base {
      * @return null|bool
      */
     public function is_enabled(): bool {
-        return !((bool) get_config('cltr_' . $this->name, 'disabled'));
+        return ((bool) get_config('cltr_' . $this->name, 'enabled'));
     }
 
     /**
@@ -86,7 +86,7 @@ class cltr extends \core\plugininfo\base {
      */
     public function set_enabled(bool $enable) {
         if ($this->is_enabled() != $enable) {
-            set_config('disabled', (int) !$enable, 'cltr_' . $this->name);
+            set_config('enabled', (int) $enable, 'cltr_' . $this->name);
             \core_plugin_manager::reset_caches();
         }
     }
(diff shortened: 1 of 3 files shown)


collector/cloudwatch/settings.php
```diff
@@ -75,6 +75,12 @@
                 get_string('environment_desc', 'cltr_cloudwatch'),
                 'Dev', PARAM_TEXT));
         } else {
+            // When tool_cloudmetrics is installed, the AWS cloudwatch plugin is installed and enabled.
+            // Disable it if this is the first visit to the settings page and it has never been configured.
+            if (! get_config('cltr_cloudwatch', 'firstvisit')) {
```
brendanheywood: Most of the collectors we might eventually ship with should start out as disabled. ie if we support azure, prometheus, GCP, AWS, most people are only going to want one and just turn on that one.

So I think we want either:

option a) a way for each collector plugin to declare if it should default on
option b) we assume they should all default off, except database which is the special case and can default on
option c) we default them all off

I think the root cause here is that if there is no config at all then it assumes it is on:

```
$ php admin/cli/cfg.php --component=cltr_database
version	2022040500
metric_expiry	2592000
```

then if you disable it it is set:

```
$ php admin/cli/cfg.php --component=cltr_database
version	2022040500
metric_expiry	2592000
disabled	1
```
then re-enable:

```
$ php admin/cli/cfg.php --component=cltr_database
version	2022040500
metric_expiry	2592000
disabled	0
```

I think the simplest is to swap this logic so it must be explicitly enabled, and they default off, and we can enable the database either in install.php or upgrade.php

https://docs.moodle.org/dev/Plugin_files#db.2Finstall.php

macgirvin: I do like that better.

macgirvin:
I'm looking into creating install.php now to set the database collector to enabled on installation.

jaypha:
I think there should be an upgrade step that would swap the config value in an existing installation. That is, see what the 'disabled' config is, and then set 'enabled' accordingly (and eliminate the 'disabled' config for good measure).

jaypha:
Also, for good measure, there should be a version bump.

Issue #26, opened 2022-03-17
readme docs on ideal setup to reduce random latency from cron contention


brendanheywood:
ie if there is any chance of cron runner contention set this task up with its own crontab entry to it always runs on a clean minute boundary and won't be delayed if other cron load is high. If the normal cron runs it then that's fine

PR #83, opened 2022-05-13
Issue #73: Use the starting time of the cron process instead of the task

### LINKED ISSUE #73: Bug with cloudwatch skipping time periods
There is a weird bug where only 2 out of 3 times end up in cloud watch so we get a discontinuous line graph

image

jaypha:
Resolves #73

Add mtrace temporarily for debugging.

### DIFF
--- a/classes/metric/builtin_user_base.php
+++ b/classes/metric/builtin_user_base.php
@@ -39,6 +39,15 @@ abstract class builtin_user_base extends builtin_base {
     }
 
     /**
+     * True if this metric is capable of generating metric items for past times.
+     *
+     * @return bool
+     */
+    public function can_generate_past_metric_items(): bool {
+        return true;
+    }
+
+    /**
      * Generates the metric items from the source data.
      *
      * Starting from $finishtime, this will generate an item for each frequency period (as defined by get_frequency()),
@@ -64,11 +73,21 @@ abstract class builtin_user_base extends builtin_base {
 
         // We go backwards because we want to align with finishtime.
         $time = $finishtime;
-        while ($time > $starttime) {
+        do { // In the special case where $starttime = $finishtime, we allow one iteration.
             $users = $DB->count_records_select('user', $selectclause, [$time - $window, $time]);
             $items[] = new metric_item($name, $time, $users, $this);
             $time = \tool_cloudmetrics\lib::get_previous_time($time, $freq);
-        }
+        } while ($time > $starttime);
         return array_reverse($items);
+    }
+
+    /**
+     * Generate a single metric item from source data using the immediate time.
+     *
+     * @return metric_item
+     */
+    public function generate_metric_item(): metric_item {
+        $ts = time();
+        return $this->generate_metric_items($ts, $ts)[0];
     }
 }
--- a/classes/metric/builtin_user_base.php
+++ b/classes/metric/builtin_user_base.php
@@ -39,55 +39,35 @@ abstract class builtin_user_base extends builtin_base {
     }
 
     /**
-     * True if this metric is capable of generating metric items for past times.
+     * Time window to draw user data from for the metric.
      *
-     * @return bool
+     * @return int Time as seconds.
      */
-    public function can_generate_past_metric_items(): bool {
-        return true;
+    protected function get_time_window(): int {
+        $value = (int) get_config('tool_cloudmetrics', $this->get_name() . '_time_window');
+        if ($value < 1) {
+            $value = 1;
+        }
+        return $value;
     }
 
     /**
      * Generates the metric items from the source data.
      *
-     * Starting from $finishtime, this will generate an item for each frequency period (as defined by get_frequency()),
-     * going backwards until $startime is reached.
-     *
-     * If the metric type is unable to generate items from past data, the parameters will be ignored, and a
-     * single item from the immediate time will be generated.
+     * Uses $starttime to $finishtime to draw from the source data.
      *
      * @param int $starttime
      * @param int $finishtime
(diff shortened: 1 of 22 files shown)


brendanheywood:
I think each metric should be passed in the window for when the metric should have been for

Ie more or less round down to the interval it should be rather than exactly what it is. I do not think we should reach into the cron managers internals

jaypha:
Wouldn't that just artificially force it to match?
If the current time is 12:31, how do we know if it is the 12:31 cron run or the 12:30 cron run?

brendanheywood:
From an implementation angle, we have some metrics which can only ever be 'now' and then we have some other that we can retrospectively work out based on things like logs. From the interface's point of view we don't quite know the difference, and I'm on the fence around whether it should know or not. No matter what when cron runs there will always be some delay, it could be seconds or it could be a minute or more. So I think when the manager asks a metric for a value it should always provide the start and end timestamps for the period it should be measuring, and the metric_item it returns should reflect this. If it is a 'now' metric it can just ignore the time period as it can't use it. But if it is something that say reads the logs then it can use it exactly, and it also means it could be used to backfill data. So there is some overlap with the work @marcghaly is doing.

```
$metric->get_metric_item($start, $finish);
```

Both timestamps will always be in the past no matter what. If cron is really laggy then we should detect this and skip some as needed. eg the metric could have internal logic like 'I am being asked for N-10 to N and it is currently N+1 and this is ok, but if it was N+5 we are closer to the next one so. That could be shared in the base class.

But lag of 1 minute when we are measuring a metric every 5 minutes (the real example here) or every hour or every 24 hours is totally fine and in these cases we want to store the nice rounded stamp of when it should have ideally run not the exact timestamp. Especially if we are graphing multiple metrics side by side we want them to all align on the same time period boundaries.

If we are measuring a metric every minute and cron is on average lagging more than that then we have a more important problem to fix. This is more or less this issue where I want to document how to set up a highly accurate crond setup for when this is needed:

https://github.com/catalyst/moodle-tool_cloudmetrics/issues/26

ie move the metric task to it's own dedicated crond entry so it isn't contending with anything else. We could also add a check for this (rainy day task, not important now as will only ship with metrics which default to 5 mins)

https://github.com/catalyst/moodle-tool_cloudmetrics/issues/86

jaypha:
This PR is now a significant refactoring.

The function get_metric_item() for metric classes is now called generate_metric_items($starttime, $finishtime).
There is a new config for storing the latest generate.
The collect metrics task has been refactored to use this config value. It should no longer be dependent on the current time being exact.

classes/task/collect_metrics_task.php
```diff
@@ -125,42 +79,36 @@ public function get_frequency_cutoff(int $timediff): int {
...
             }
-        }
 
-        $metrics = metric\manager::get_metrics(true);
-        $items = [];
-        foreach ($metrics as $metric) {
-            if (($metric->get_frequency() <= $cutoff || $metric->get_frequency() == $ismonth) && $metric->is_ready()) {
-                $items[] = $metric->get_metric_item();
+            if ($duetimestamp <= $nowtimestamp) {
+                mtrace('generating metric for ' . $metrictype->get_name() . ' from ' . userdate($lasttimestamp) . ' to ' . userdate($duetimestamp));
```
brendanheywood: 1) This is looking nice - can we do this in a slightly different place so we can include the metric itself?

2) Also can use do some sprintf alignment so it looks like a text table.

3) tighten up the days and months just to make it a little shorter

```
generating metric for activeusers    13 from Wed, 1 Jun 2022, 11:53 AM to Wed, 1 Jun 2022, 11:54 AM
generating metric for newusers        6 from Wed, 1 Jun 2022, 11:53 AM to Wed, 1 Jun 2022, 11:54 AM
generating metric for onlineusers     8 from Wed, 1 Jun 2022, 11:53 AM to Wed, 1 Jun 2022, 11:54 AM

```

tests/tool_cloudmetrics_collect_metrics_test.php
```diff
@@ -34,7 +34,7 @@ public function receive(array $names) {
 
 /**
  * A class to help test the collect metrics task. This class overrides collect_metrics_task
- * so that instead of sending th emetric items to the colletors, it passes it to a
+ * so that instead of sending th emetric items to the collectors, it passes it to a
```
brendanheywood: unrelated typo  'th emetric'

classes/metric/builtin_user_base.php
```diff
@@ -0,0 +1,74 @@
...
+     * going backwards until $startime is reached.
+     *
+     * If the metric type is unable to generate items from past data, the parameters will be ignored, and a
+     * single item from the immediate time will be generated.
+     *
+     * @param int $starttime
+     * @param int $finishtime
+     * @return array An array of metric items, in forward chronological order.
+     */
+    public function generate_metric_items(int $starttime, int $finishtime): array {
```
brendanheywood: I'm a little confused around the semantics of this one. We need to clearly split metrics which support only live-ish / now type metrics from metrics which can retro back fill data as needed. It looks like this is attempting to do both?

The bug we just talked about, IF a metric does support back filling then we could just ask it to do a small batch of back filling if we missed a few cron ticks, but we can do that as bonus points after we get the basics sorted.

jaypha:
I've reseparated the generate functions and added a can_generate_past_metric_items() function to the metric base.
I don't want to go too deep into supporting backfilling as it is not really in the scope of this PR.
I added a second mtrace for printing info about items, so that if a crash occurs during generate_metric_items, we at least have a clue as to what it was trying to do.

Issue #67, opened 2022-05-05
Have an option for a metric to declare that it can populate retrospectively to backfill data


brendanheywood:
If this is true for a metric then we probably want the interface for this to be quite different where we pass into the metric the time resolution (eg 5 mins) and the time period start and end date to backfill OR a max number of samples we want (or both), and it does all the logic and passes back the metrics which then the manager sends to the collectors.

Reason for this is it is going to be way more efficient this way.

Then wrap this in a new setting so we can also force set it, which is how back back in tie we should back fill. The manager is probably going to need to manage state on behalf of all the metrics, or each metric needs to be able to say what state it already has.

Need to then make some business decisions around which collectors support this, I would strongly assume that only the internal db collector will at first.

brendanheywood:
Actually in hindsite it is probably better to simply pass in the minimum time period to the metric, and then each internally handles all it's own logic so say if it needs to backfill or not. This will be idempotent and future runs will do nothing. This backfill can be run very infrequently like nightly.

marcghaly:
Hi @brendanheywood,

I was wondering if there was a reason `usort` was used instead of `uasort` here :

https://github.com/catalyst/moodle-tool_cloudmetrics/blob/MOODLE_35_STABLE/classes/metric/manager.php#L117

This is provoking a moodle exception to be thrown. 

Cheers,

Marc

brendanheywood:
No reason, but it isn't failing for us so far. In any case we are expecting rough edges as this is all hot off the press and not battle tested so just make issues and fix them as you see fit

jaypha:
Actually, this is what caused issue #71.

marcghaly:
Hi @brendanheywood,

Concerning the display of those informations, should it be accessible from this panel as well ?

![Screenshot from 2022-05-05 17-26-46](https://user-images.githubusercontent.com/76186988/167499683-c4ece381-a48a-4d66-8675-011e519cf442.png)

Should we use a form? Is there some specific concerning retrieving data from `logstore_standard_log` table?

Cheers,

Marc

brendanheywood:
@marcghaly a rough outline of how I see this going:

1) a metric should declare that it can be populated retrospectively. Not all metrics can be as they only make sense in the 'now' and we won't have a specific log table or any other table

2) even if a metric can be back filled retrospectively doesn't mean we want it to be. That would be an admin option to specify how far back it should go

[...]

PR #92, opened 2022-05-26
Issue #67: Backfilled metric - online users

marcghaly:
Hi @jaypha, hi @brendanheywood, 

This is PR for issue #67.

Regards,

Marc

### DIFF
--- a/classes/admin_setting_manage_metrics.php
+++ b/classes/admin_setting_manage_metrics.php
@@ -85,7 +85,7 @@ class admin_setting_manage_metrics extends \admin_setting {
         $txt->frequency = get_string('frequency', 'tool_cloudmetrics');
 
         $table = new \html_table();
-        $table->head  = array($txt->plugin, $txt->name, $txt->description, $txt->frequency, $txt->actions);
+        $table->head  = array($txt->plugin, $txt->name, $txt->description, $txt->frequency, $txt->actions, get_string('backfillable', 'tool_cloudmetrics'));
         $table->align = array('left', 'left', 'left', 'left');
         $table->attributes['class'] = 'manageformattable generaltable admintable w-auto';
         $table->data  = array();
@@ -133,6 +133,15 @@ class admin_setting_manage_metrics extends \admin_setting {
                 $attributes
             );
 
+            $backfillurl = new \moodle_url('/admin/tool/cloudmetrics/collector/database/backfill.php', ['metric' => $metric->get_name()]);
+            // Metric backfill support and if so - link.
+            if ($metric->is_backfillable()) {
+                $class = '';
+                $backfilllink = \html_writer::link($backfillurl->out(false), $metric->get_label() . ' backfill');
+            } else {
+                $backfilllink = get_string('no_support', 'tool_cloudmetrics');
+            }
+
             // Inplace editables required an integer ID, whereas metrics are identified with strings.
             // We use a hash function to convert to an integer, to future proof the code.
             $intid = hexdec(substr(md5($metric->get_name()), 0, 8));
@@ -155,7 +164,8 @@ class admin_setting_manage_metrics extends \admin_setting {
                 $displayname,
                 $description,
                 $OUTPUT->render($editable),
-                $hideshow . $settingslink . $chartlink
+                $hideshow . $settingslink . $chartlink,
+                $backfilllink,
             ]);
             if ($class) {
                 $row->attributes['class'] = $class;
--- a/classes/collector/base.php
+++ b/classes/collector/base.php
@@ -56,4 +56,13 @@ abstract class base {
     public function is_ready(): bool {
         return true;
     }
+
+    /**
+     * Abilitity for a collector to save old data.
+     *
+     * @return bool
+     */
+    public function supports_backfillable_metrics(): bool {
+        return false;
+    }
 }
--- a/classes/metric/base.php
+++ b/classes/metric/base.php
@@ -126,6 +126,30 @@ abstract class base {
     }
 
     /**
+     * The range for which a backfilled metric has retrieved data.
+     *
+     * @return array|bool First value is min timestamp, second is max timestamp, third is used interval.
+     * @throws \dml_exception
+     */
+    public function get_range_retrieved() {
+        $config = explode('-', get_config('tool_cloudmetrics', $this->get_name() . '_range'));
+        if (count($config) !== 3) {
+            return [-1, -1, -1];
+        }
+        return [(int)$config[0], (int)$config[1], (int)$config[2]];
+    }
+
+    /**
+     * Sets the range for which a backfilled metric has retrieved data.
+     *
+     * @param array $currentconfig Config containing min and max timestamps plus interval.
+     */
+    public function set_range_retrieved(array $currentconfig) {
+        $currentconfig = implode('-', $currentconfig);
+        set_config($this->get_name() . '_range', $currentconfig, 'tool_cloudmetrics');
+    }
+
+    /**
      * The latest time for which a metric item was generated for.
      *
      * @return int
@@ -154,4 +178,38 @@ abstract class base {
      * @return metric_item
      */
     abstract public function generate_metric_item(int $starttime, int $finishtime): metric_item;
+
+    /**
+     * Metric's ability to be backfilled.
+     *
+     * @return bool
+     */
+    public function is_backfillable(): bool {
+        return false;
+    }
+
+    /**
+     * Returns records for backfilled metric.
+     *
+     * @param int $backwardperiod Time from which sample is to be retrieved.
+     * @param int $finishtime If data is being completed argument is passed here.
+     *
+     * @return array|null
+     */
+    public function generate_metric_items(int $backwardperiod, int $finishtime = null): ?array {
+        return null;
+    }
+
+    /**
+     * Returns two dates representing start and end log points, hence available data to retrieve.
+     *
+     * @return object|bool
+     */
+    public function get_range_log_available() {
+        global $DB;
+
+        $sql = 'SELECT max(timecreated) AS "max", min(timecreated) AS "min"
+                  FROM {logstore_standard_log}';
+        return $DB->get_record_sql($sql);
+    }
 }
--- a/classes/metric/online_users_metric.php
+++ b/classes/metric/online_users_metric.php
@@ -29,6 +29,18 @@ class online_users_metric extends builtin_user_base {
     /** @var string The DB field the metric accesses. */
     protected $dbfield = 'lastaccess';
 
+    /** @var int The interval config for which data is displayed in seconds (eg: 5 minutes = 300). */
+    public $interval;
+
+    /** @var int The minimal timestamp representing earliest date retrieved in DB. */
+    public $mintimestamp;
+
+    /** @var int The maximal timestamp representing latest date retrieved in DB. */
+    public $maxtimestamp;
+
+    /** @var bool True if config used is same as one requested. */
+    public $sameconfig;
+
     /**
      * The metric's name.
      *
@@ -36,5 +48,102 @@ class online_users_metric extends builtin_user_base {
      */
     public function get_name(): string {
         return 'onlineusers';
+    }
+
+    /**
+     * Metric's ability to be backfilled.
+     *
+     * @return bool
+     */
+    public function is_backfillable(): bool {
+        return true;
+    }
+
+
+    /**
+     * Returns records for backfilled metric.
+     *
+     * @param int $backwardperiod Time from which sample is to be retrieved.
+     * @param int $finishtime If data is being completed argument is passed here.
+     *
+     * @return array
+     */
+    public function generate_metric_items($backwardperiod, $finishtime = null): array {
+        global $DB;
+
+        // Get start time from period selection.
+        $starttime = time() - $backwardperiod;
+        $finishtime = ($finishtime === -1) ? null : $finishtime;
+        // Allows data to be completed instead of retrieving all data again unless frequency change.
+        $finishtime = $finishtime ?? time();
+        $frequency = $this->get_frequency();
+        [$mintmptmp, $maxtmpstmp, $freqretrieved] = $this->get_range_retrieved();
+
+        if ($finishtime < $starttime) {
+            return [];
+        }
+        $secondsinterval = [
+            manager::FREQ_MIN => MINSECS,
+            manager::FREQ_5MIN => MINSECS * 5,
+            manager::FREQ_15MIN => MINSECS * 15,
+            manager::FREQ_30MIN  => MINSECS * 30,
+            manager::FREQ_HOUR => HOURSECS,
+            manager::FREQ_3HOUR => HOURSECS * 3,
+            manager::FREQ_12HOUR => HOURSECS * 12,
+            manager::FREQ_DAY => DAYSECS,
+            manager::FREQ_WEEK => WEEKSECS,
+            manager::FREQ_MONTH => WEEKSECS * 4
+        ];
+
+        $interval = $secondsinterval[$frequency];
+        $sql = 'SELECT floor(timecreated / :interval ) * :intervaldup AS time,
+                       COUNT(DISTINCT userid) as value
+                  FROM {logstore_standard_log}
+                 WHERE timecreated >= :starttime
+                   AND timecreated <= :finishtime
+              GROUP BY 1
+              ORDER BY 1 ASC';
+        $rs = $DB->get_recordset_sql($sql,
+                ['interval' => $interval, 'intervaldup' => $interval, 'starttime' => $starttime, 'finishtime' => $finishtime]);
+        $metricitems = [];
+        $count = 0;
+        foreach ($rs as $r) {
+            if ($count !== 0 && $metricitems[$count - 1]->time + $interval !== (int)$r->time) {
+                // Code to add times where no user have been concurrently active.
+                for ($i = $metricitems[$count - 1]->time + $interval; $i <= $r->time; $i += $interval) {
+                    if ($i === (int)$r->time) {
+                        $metricitems[] = new metric_item($this->get_name(), $r->time, $r->value, $this);
+                    } else {
+                        $metricitems[] = new metric_item($this->get_name(), $i, 0, $this);
+                        $count++;
+                    }
+                }
+            } else {
+                $metricitems[] = new metric_item($this->get_name(), $r->time, $r->value, $this);
+            }
+            $count++;
+        }
+        $rs->close();
+
+        $this->mintimestamp = !empty($metricitems) ? $metricitems[0]->time : null;
+        $this->maxtimestamp = !empty($metricitems) ? end($metricitems)->time : null;
+        $this->interval = $frequency;
+
+        return $metricitems;
+
+    }
+
+    /**
+     * Stores what data has been sent to collector.
+     *
+     */
+    public function set_data_sent_config() {
+        // Store what data has been sent min, max timestamp range and interval.
+        $currentconfig = [$this->mintimestamp, $this->maxtimestamp, $this->interval];
+        $rangeretrieved = $this->get_range_retrieved();
+        $this->sameconfig = ($rangeretrieved === $currentconfig);
+        if (!$this->sameconfig && isset($this->mintimestamp) && isset($this->maxtimestamp) && isset($this->interval)) {
+            $this->set_range_retrieved($currentconfig);
+        }
     }
 }
--- a/collector/database/chart.php
+++ b/collector/database/chart.php
@@ -14,6 +14,11 @@
 // You should have received a copy of the GNU General Public License
 // along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
 
+use cltr_database\form\metric_backfill_form;
+use cltr_database\collector;
+use core\chart_line;
+use core\chart_series;
+use single_select;
 use tool_cloudmetrics\metric;
 
 /**
@@ -51,6 +56,10 @@ if ($defaultperiod === -1) {
 }
 
 $metrics = metric\manager::get_metrics(true);
+// Error management if metric is not enabled.
+if (!isset($metrics[$metricname])) {
+    throw new moodle_exception('metric_not_enabled', 'tool_cloudmetrics', '', $metricname);
+}
 $metriclabels = [];
 foreach ($metrics as $m) {
     $metriclabels[$m->get_name()] = $m->get_label();
@@ -59,16 +68,16 @@ foreach ($metrics as $m) {
     }
 }
 
-$select = new \single_select(
+$select = new single_select(
     $url,
     'metric',
     $metriclabels,
     $metricname
 );
 $select->set_label(get_string('select_metric_for_display', 'cltr_database'));
+$context = [];
 
 // Prepare time window selector.
-
 $periods = [
     HOURSECS      => get_string('one_hour', 'tool_cloudmetrics'),
     DAYSECS       => get_string('one_day', 'tool_cloudmetrics'),
@@ -82,6 +91,8 @@ $periods = [
     YEARSECS * 2  => get_string('two_year', 'tool_cloudmetrics'),
 ];
 
+$collector = new \cltr_database\collector();
+
 // Create a new URL object to avoid poisoning the existing one.
 $url = clone $url;
 $url->param('metric', $metricname);
@@ -91,11 +102,12 @@ $periodselect = new \single_select(
     $periods,
     $defaultperiod
 );
+
+$backfillurl = new moodle_url('/admin/tool/cloudmetrics/collector/database/backfill.php', ['metric' => $metricname]);
+
 $periodselect->set_label(get_string('select_graph_period', 'cltr_database'));
 
-$collector = new \cltr_database\collector();
 $records = $collector->get_metrics($metricname, $defaultperiod);
-
 $values = [];
 $labels = [];
 
@@ -104,19 +116,22 @@ foreach ($records as $record) {
     $labels[] = userdate($record->time, get_string('strftimedatetime', 'cltr_database'), $CFG->timezone);
 }
 
-$chartseries = new \core\chart_series($metriclabels[$metricname], $values);
-
-$chart = new \core\chart_line();
+$chartseries = new chart_series($metriclabels[$metricname], $values);
+$chart = new chart_line();
 $chart->add_series($chartseries);
 $chart->set_labels($labels);
 
+$context['chart'] = $OUTPUT->render($chart);
+$context['selector'] = $OUTPUT->render($select);
+$context['periodselect'] = $OUTPUT->render($periodselect);
+$context['backfillurl'] = $backfillurl;
+$context['backfillable'] = $metrics[$metricname]->is_backfillable();
+$context['metriclabel'] = $metric->get_label();
+$context['metricdescription'] = $metrics[$metricname]->get_description();
+$context['metriclabeltolower'] = strtolower($metric->get_label());
+
+$renderer = $PAGE->get_renderer('tool_cloudmetrics');
+
 echo $OUTPUT->header();
-echo $OUTPUT->render($select);
-echo html_writer::empty_tag('br');
-echo $OUTPUT->render($periodselect);
-if (isset($metric)) {
-    echo html_writer::tag('h3', $metric->get_label());
-    echo html_writer::tag('p', $metric->get_description());
-}
-echo $OUTPUT->render($chart);
+echo $renderer->render_chart_page($context);
 echo $OUTPUT->footer();
--- a/collector/database/classes/collector.php
+++ b/collector/database/classes/collector.php
@@ -28,8 +28,12 @@ use tool_cloudmetrics\metric;
  * @copyright 2022, Catalyst IT
  * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
-
 class collector extends base {
+    /**
+     * Records metric in cltr table.
+     *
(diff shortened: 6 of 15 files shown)


classes/metric/concurrent_users_metrics.php
```diff
@@ -0,0 +1,134 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_cloudmetrics\metric;
+
+use cltr_database\collector;
+
+/**
+ * Metric class for new users.
```
jaypha: Need to describe this class.

classes/metric/active_users_metric.php
```diff
@@ -55,4 +55,13 @@ public function get_metric_item(): metric_item {
         $users = $DB->count_records_select('user', 'lastlogin > ?', [$now - $this->get_time_window()]);
         return new metric_item($this->get_name(), $now, $users, $this);
     }
+
+    /**
+     * Metric's ability to be backfilled.
+     *
+     * @return bool
+     */
+    public function is_backfillable(): bool {
```
jaypha: If this is going to appear in every metric, then it should also be put into the base class, even if it is abstract. Although it should probably default to returning false.

classes/metric/concurrent_users_metrics.php
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Returns records for backfilled metric.
+     *
+     * @param array $data Array of data specifying timeframe and sample to backfill.
+     * @return array|null
+     */
+    public function get_backfilled_metric(array $data): ?array {
+        global $DB;
+        $collector = new collector();
```
jaypha: Metrics are not supposed to be aware of collectors. Especially specific ones. You should be able to match any collector to any metric and have it work.

classes/metric/concurrent_users_metrics.php
```diff
@@ -0,0 +1,134 @@
...
+        return true;
+    }
+
+    /**
+     * Returns records for backfilled metric.
+     *
+     * @param array $data Array of data specifying timeframe and sample to backfill.
+     * @return array|null
+     */
+    public function get_backfilled_metric(array $data): ?array {
```
jaypha: This should also be declared in the base class, with a default of doing nothing.

collector/database/classes/collector.php
```diff
@@ -39,19 +45,41 @@ public function record_metric(metric_item $item) {
...
      * @throws \dml_exception
      */
-    public function get_metrics($metricnames = null, $since = false): array {
+    public function get_metrics($metricnames = null, $metricclass = null, $since = false, $data = []): array {
         global $DB;
         $starting = '';
+
+        if (!is_null($metricclass)) {
+            if ($metricclass->is_backfillable()) {
+                $backfilleddata = $metricclass->get_backfilled_metric($data);
```
jaypha: Is this the best place for it?
This function (which I now realise is poorly named. It should be called 'get_saved_metric_data', or something similar), is about retrieving saved metric data, not about generating them. Perhaps this should be done where metric data is originally generated.

This function is unique to cltr_database\collector. Other collectors may have the ability to record backfill data, but not have the ability to retrieve it (like AWS).

collector/database/classes/form/metric_backfill_form.php
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Form for specifying user metric data to be backfilled.
+ *
+ * @package    cltr_database
+ * @copyright  2022 Catalyst IT Australia Pty Ltd
+ * @author     Ghaly Marc-Alexandre <marc-alexandreghaly@catalyst-ca.net>
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
```
jaypha: Excess whitespace.

jaypha:
There are some issues here which I have concerns about. I'll be happy to discuss if you want to.

classes/metric/concurrent_users_metrics.php
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * Metric class for new users.
+ *
+ * @package    metric_foobar
+ * @author     Marc-Alexandre Ghaly <marc-alexandreghaly@catalyst-ca.net>
+ * @copyright  2022, Catalyst IT
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class concurrent_users_metrics extends builtin {
```
jaypha: Isn't this the same as online_users_metric?

classes/metric/concurrent_users_metrics.php
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    /**
+     * Retrieves the metric.
+     *
+     * @param  $backfilledrecord If data is backfilled, record is passed here.
+     * @return metric_item
+     */
+    public function get_metric_item($record = null): metric_item {
+        return new metric_item($this->get_name(), $record->time, $record->value, $this);
```
jaypha: How is this supposed to work during normal metric gathering?
What happens when $record is null?

classes/metric/concurrent_users_metrics.php
marcghaly: Fixed

classes/metric/active_users_metric.php
marcghaly: Fixed

classes/metric/concurrent_users_metrics.php
marcghaly: I have split and renamed this function, please let me know if this suits.

classes/metric/concurrent_users_metrics.php
marcghaly: Fixed

collector/database/classes/collector.php
marcghaly: I have split this function with others - as stated in #67, I also added a declaration whether or not a collector supports metric recording - should there also be a declaration as whether they can be retrieved or not ?

collector/database/classes/form/metric_backfill_form.php
marcghaly: Fixed

classes/metric/concurrent_users_metrics.php
marcghaly: I understood that online users are users in the sense of right now whereas concurrent users are values from the past, if those are the same I will remove the extra one.

classes/metric/concurrent_users_metrics.php
marcghaly: If possible I would like to discuss this in a bbb maybe your Monday or Tuesday morning (Sunday or Monday evening for me) ?

jaypha: Tues morning (Aussie time). If you can specify a specific time, I can make sure I am here for it. Any time is fine except for the 10:00 meeting.

jaypha: That is, I have meetings between 10:00 and 10:30 (Sydney time). Any other time is fine.

classes/metric/concurrent_users_metrics.php
jaypha: In this case the online_users_metric is redundant as concurrent_users_metric can fill both roles.

collector/database/classes/collector.php
```diff
@@ -72,4 +89,33 @@ public function get_metrics($metricnames = null, $since = false): array {
...
         return $DB->get_records_sql($sql, $params);
     }
+
+    /**
+     * Records retrieved data in collector.
+     *
+     * @param  object $metricclass Class representing metric.
+     * @param  array $data Array of metric items.
+     */
+    public function record_saved_metrics(object $metricclass, array $metricitems = []) {
```
jaypha: 'object' as a type hint is not supported in older versions of PHP. It should be '\tool_cloudmetrics\metric\base' instead.

jaypha:
Hi Marc.

I don't think a meeting is necessary. You have addressed my primary concerns for now. I'm a little iffy about having it controlled from within the database collector, but that's something I'd rather leave to Brendan.

Just that one fix causing the CI problem, and then I am OK to merge. Get Brendan to have a look at the UI aspect.

Regards

classes/metric/concurrent_users_metrics.php
brendanheywood: +1, this should just be back filling the online_users_metric

classes/metric/concurrent_users_metrics.php
brendanheywood: hi @marcghaly 

I have a few concerns around the public interface, and there is a lot of overlap with issue #73 which @jaypha was working on here:

https://github.com/catalyst/moodle-tool_cloudmetrics/pull/83

That pr isn't ready but what I've proposed just now with Jason is that the public interface for normal metrics is changed to something 

get_metric_item($starttime, $finishtime, $periodlength)

This is because in reality all metrics are retrospective, they just might be a few seconds later due to the latency of running each cron process and we need to know the exact timeperiod that the data should be have been for (even if the metric itself ignores it and just gets 'now')

Once that is in place the retrospective metrics should end up very similar something like:

get_metric_items($starttime, $finishtime, $periodlength)

The only difference is that the start time and finish time will be some distant past timestamps and we are expecting an array of them back instead of just one. We also need to consider the problem when the number of metric items returned is vast, and we might need to limit it and return what the time period was that we are returning data for which will be a subset of what was requested. Then it can be iteratively called to progressively backfill data in smaller chunks over time and the calling code would need to manage what has been processed or not already.

So I think we should land the first PR first so you don't have a moving target. Jason will get onto that in the next day or say.

I would also still like to meet, ideally all 3 of us if we can make it work, half as a meet and greet and also to iron out any last bits on this. I'll add a couple other comments in the pr here but many of them will be redundant when this gets rewritten on top of the other pr

classes/metric/concurrent_users_metrics.php
```diff
@@ -0,0 +1,146 @@
...
+            $interval = $secondsinterval[$data['interval']] ?? MINSECS * 5;
+            $sample = $sampleoptions[$data['sample']] ?? 10;
+        } else {
+            // If form is empty we do not want to backfill anything.
+            return null;
+        }
+
+        $sql = "SELECT floor(timecreated / $interval ) * $interval AS time,
+                       COUNT(DISTINCT userid) as value
+                  FROM {logstore_standard_log}";
```
brendanheywood: We should not be querying directly from logstore_standard_log, there is an api for querying the logs instead:

https://github.com/moodle/moodle/blob/master/admin/tool/log/classes/log/manager.php#L93-L102

classes/metric/base.php
```diff
@@ -131,4 +131,23 @@ public function get_settings_url() {
...
+        return false;
+    }
+
+    /**
+     * Returns records for backfilled metric.
+     *
+     * @param array $data Array of data specifying timeframe and sample to backfill.
+     * @return array|null
+     */
+    public function get_saved_metric_data(array $data): ?array {
```
brendanheywood: Can I suggest a name tweak like get_metric_items or get_metric_items_for_period?

I also dislike the $data argument, this would be much more readable as explicit arguments

classes/metric/concurrent_users_metrics.php
marcghaly: I have removed concurrent_users_metric class.

collector/database/classes/collector.php
marcghaly: Fixed.

classes/metric/concurrent_users_metrics.php
marcghaly: Hi @brendanheywood,

I have used the API, split recordset into chunks - but it seems using the API is consuming a lot of memory, any suggestion on this ?

https://github.com/catalyst/moodle-tool_cloudmetrics/blob/MOODLE_35_STABLE-67/classes/metric/online_users_metric.php#L122

classes/metric/base.php
marcghaly: Fixed

marcghaly:
Hi @brendanheywood,

I have made changes for this PR, a concern that I have is the use of the API for retrieving logs that does not provide a faster / lighter way to retrieve concurrent users (ie: using DISTINCT in a SELECT statement - retrieving all data `*` with the 2 methods provided by the API `get_events_select_iterator`, `get_events_select` seems super heavy to me, please let me know if there is a way I am not aware of) - it seems it is taking a lot of time for this operation, if that were to be left as is, would it need a spinner display ? This is rebased against latest MOODLE_35_STABLE. 
Concerning a meeting, I am available at 8 PM EST so 10 AM AEST your time, this evening and tomorrow if you want.

Regards,

Marc

brendanheywood:
hi @marcghaly 

If the log api is too slow then do it directly like it was. But just to be be clear @jaypha was doing  / has just done a huge refactor of the way the metrics work and I didn't want to waste your time as it would affect your patch. I suspect you will need to have a look at what has landed and rewrite a lot of it (despite there being no conflicts)

We have our standup at 10, so lets chat right after that. Can you please ping me on matrix, I pinged you the other day but you didn't see it and I'm not sure I have the real 'you' in matrix

marcghaly:
Hi @brendanheywood,

I have made the following changes:

- Used `metric_item class` directly in `generate_metric_items` instead of re-calling `generate_metric_item`.
- Added the UI on another page with one input representing period.
- Displayed what data can be retrieved - and what data has already been retrieved.
- Used the interval from the `Frequency` setting in `Manage Metrics` panel.
- Added a link to backfill support in the `Manage Metrics` panel.
- Queried the logstore directly for performance purposes, so it supports mariadb and pgsql.
- Added a renderer/mustache template for chart page.
- Instead of erasing data present in `cltr` db, when date is further than one already present, only complete data.
- If frequency is changed, data would be erased - it seems confusing having different intervals displayed - can change though.

Let me know if this suits.

Regards,

Marc

brendanheywood:
hi @marcghaly 

This is looking great and pretty close, one issue I've found is that when you backfill data you don't end up with nice time periods which align with time period boundaries:

image

If we had a setting that was supposed to have been every hour then I'd expect the backfilled data to be on the hour every hour

A few other small comments in the pr

classes/admin_setting_manage_metrics.php
```diff
@@ -155,7 +164,8 @@ public function output_html($data, $query='') {
                 $displayname,
                 $description,
                 $OUTPUT->render($editable),
-                $hideshow . $settingslink . $chartlink
+                $hideshow . $settingslink . $chartlink,
+                $backfilllink
```
brendanheywood: a super small general thing, with arrays it is nicer to add the trialing , on the last item so that next time we add an entry you don't get a diff which modifies an unrelated line

classes/collector/base.php
```diff
@@ -56,4 +56,13 @@ public function record_metrics(array $metrics) {
     public function is_ready(): bool {
         return true;
     }
+
+    /**
+     * Abilitity for a collector to save old data.
+     *
+     * @return bool
+     */
+    public function supports_saved_metric_record(): bool {
```
brendanheywood: The method below is called is_backfillable, I think we should be consistent with the names in code and comments, eg something like:

is_backfillable()
supports_backfillable_metrics()

jaypha:
Hi Marc

> If we had a setting that was supposed to have been every hour then I'd expect the backfilled data to be on the hour every hour

In case you didn't know, there are several functions in the classes/lib.php which can help with this.

classes/metric/base.php
```diff
@@ -154,4 +177,37 @@ public function set_last_generate_time(int $timestamp) {
...
+
+    /**
+     * Returns two dates representing start end end log points, hence available data to retrieve.
+     *
+     * @return object|bool
+     */
+    public function get_range_log_available() {
+        global $DB;
+
+        $sql = 'SELECT max(timecreated) AS "max", min(timecreated) AS "min" FROM {logstore_standard_log}';
```
brendanheywood: would be slightly nicer to format this
https://docs.moodle.org/dev/SQL_coding_style#Indentation

classes/metric/online_users_metric.php
```diff
@@ -37,4 +49,99 @@ class online_users_metric extends builtin_user_base {
     public function get_name(): string {
         return 'onlineusers';
     }
+
+    /**
+     * The metric type.
+     *
+     * @return int
+     */
+    public function get_type(): int {
```
brendanheywood: Is this needed in this patch? Seems unrelated and isn't used elsewhere in the patch?

classes/metric/base.php
```diff
@@ -154,4 +177,37 @@ public function set_last_generate_time(int $timestamp) {
...
+
+    /**
+     * Returns records for backfilled metric.
+     *
+     * @param int $backwardperiod Time from which sample is to be retrieved.
+     * @param int $finishtime If data is being completed argument is passed here.
+     *
+     * @return array|null
+     */
+    public function generate_metric_items(int $backwardperiod, int $finishtime = null): ?array {
```
brendanheywood, line 199: One thing I would like to clear up in the contract for a backfillable metric and the php docs etc is the alignment of dates for the period and finish time.

For instance, lets say you have a metric which is run every day, but we only started today. We want to back fill it for a year. Right now it is midday and we start by back-filling it for just a week. We need to be clear about what the starting point is for the period and weather this is 100% the responsibility of the metric to calculate correctly.

Because if we ask for 1 week of data, but going back from now at midday then we will end up with half a days data left over at the other end. We could end up saving an incorrect metric because its only counting part of the data for that period. The same is true of the $finishtime. I don't mind which side of the fence it is on as long as it works and is clearly documented, but it is probably better shared as much as possible and done on the collector chart side. ie $finishtime will always be exactly on a boundary of a frequency time slice, and $backwardperiod will always be some integer product of the frequency.

I also think there will be subtle bugs when it comes to daylights savings but I think we can just ignore it and live with the slight errors. At least all the data will appear in one metric or the adjacent one so it should mostly balance out.

classes/metric/online_users_metric.php
```diff
@@ -37,4 +49,99 @@ class online_users_metric extends builtin_user_base {
...
+            manager::FREQ_12HOUR => HOURSECS * 12,
+            manager::FREQ_DAY => DAYSECS,
+            manager::FREQ_WEEK => WEEKSECS,
+            manager::FREQ_MONTH => WEEKSECS * 4
+        ];
+
+        $interval = $secondsinterval[$frequency];
+        $sql = 'SELECT floor(timecreated / :interval ) * :intervaldup AS time,
+                       COUNT(DISTINCT userid) as value
+                  FROM {logstore_standard_log} WHERE timecreated >= :starttime AND timecreated <= :finishtime
```
brendanheywood: a few line breaks with sql indents here

classes/metric/online_users_metric.php
```diff
@@ -37,4 +49,99 @@ class online_users_metric extends builtin_user_base {
...
+    }
+
+    /**
+     * Stores what data has been sent to collector.
+     *
+     */
+    public function set_data_sent_config() {
+        // Store what data has been sent min, max timestamp range and interval.
+        $currentconfig = $this->mintimestamp . '-' . $this->maxtimestamp . '-' . $this->interval;
+        $this->sameconfig = ($this->get_range_retrieved() === $currentconfig) ? true : false;
```
brendanheywood: should not need the ternary at the end

collector/database/chart.php
```diff
@@ -59,16 +65,57 @@
...
     $metriclabels,
     $metricname
 );
 $select->set_label(get_string('select_metric_for_display', 'cltr_database'));
+$context = [];
+// Error management if metric is not enabled.
+if (!isset($metrics[$metricname])) {
+    throw new moodle_exception('metric_not_enabled', 'tool_cloudmetrics', '', $metricname);
+}
+$backfilledmetric = $metrics[$metricname]->is_backfillable();
```
brendanheywood: This page is about showing the chart, I'd prefer it if all the logic around backfilling was moved to a new endpoint which you post to, and then after it has filled the data then it redirects back to the chart page as a get.

marcghaly:
Hi @brendanheywood,

Just checking, in the following example it is set to be hourly, for example between two first rows, should it display between 12:00 and 16:00 (12, 13, 14, 15 , 16) with a 0 value ? Is that what you mean ? I am using pgsql maybe issue is related to mariadb/mysqli 

[](url)
Screenshot from 2022-06-07 09-39-17

This is what I get with mysql:
Screenshot from 2022-06-07 09-57-47


Regards,

Marc

collector/database/chart.php
```diff
@@ -59,16 +65,57 @@
...
+    $daterange = $metrics[$metricname]->get_range_log_available();
+    $startdate = userdate($daterange->min, get_string('strftimedatetime', 'cltr_database'), $CFG->timezone);
+    $enddate = userdate($daterange->max, get_string('strftimedatetime', 'cltr_database'), $CFG->timezone);
+    $mform = new metric_backfill_form(null, $daterange);
+    $context['form'] = $mform->render();
+    if ($fromform = $mform->get_data()) {
+        $periodretrieval = $fromform->periodretrieval;
+        $finishtime = $datametricsaved[0] ?? null;
+        $metricitems = $metrics[$metricname]->generate_metric_items($periodretrieval, $finishtime);
+        if ($collector->supports_saved_metric_record()) {
```
brendanheywood: I think check this much earlier at the top and throw an error if it is not backfillable

classes/metric/base.php
```diff
@@ -125,6 +125,29 @@ public function get_settings_url() {
         return null;
     }
 
+    /**
+     * The range for which a backfilled metric has retrieved data.
+     *
+     * @return string|bool
+     * @throws \dml_exception
+     */
+    public function get_range_retrieved() {
```
brendanheywood, line 134: would it be nicer to the calling code if this returned an array of two values? would remove a lot of explodes around the place

collector/database/classes/collector.php
```diff
@@ -72,4 +88,40 @@ public function get_metrics($metricnames = null, $since = false): array {
...
+     * @param  \tool_cloudmetrics\metric\base $metricclass Class representing metric.
+     * @param  array $data Array of metric items.
+     */
+    public function record_saved_metrics(\tool_cloudmetrics\metric\base $metricclass, array $metricitems = []) {
+        global $DB;
+
+        $maxbackdate = explode('-', $metricclass->get_range_retrieved())[0];
+        $transaction = $DB->start_delegated_transaction();
+        // Sets what data has been sent to collector.
+        $metricclass->set_data_sent_config();
```
brendanheywood: you should avoid setting config inside a transaction, if it rolls back you can end up with the db and the MUC out of sync because the MUC doesn't roll back. I'd do it after it has committed

collector/database/classes/collector.php
```diff
@@ -72,4 +88,40 @@ public function get_metrics($metricnames = null, $since = false): array {
...
+        if (is_array($metricitems) && count($metricitems) != 0 && !$metricclass->sameconfig) {
+            // When a further date is passed we only want to add missing data.
+            $iscompleted = ($metricitems[0]->time < $maxbackdate) ? true : false;
```
brendanheywood: again ternary is redundant

collector/database/classes/collector.php
```diff
@@ -72,4 +88,40 @@ public function get_metrics($metricnames = null, $since = false): array {
...
+            if ($iscompleted) {
+                // Trick to avoid erasing data when it only needs to be completed.
+                $this->record_metrics($metricitems);
+            } else {
+                $this->delete_metrics($metricclass->get_name());
```
brendanheywood: I'm a little confused why we'd ever need to delete data? If we are being asked to backfill data from time period which already has data then that smells like what ever is calling record_saved_metrics isn't sending correct times

collector/database/classes/form/metric_backfill_form.php
```diff
@@ -0,0 +1,65 @@
...
+ * @author    Ghaly Marc-Alexandre <marc-alexandreghaly@catalyst-ca.net>
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class metric_backfill_form extends moodleform {
+    public function definition() {
+        $mform = $this->_form;
+        $maxbackwardsdate = $this->_customdata->min;
+        $maxperiod = time() - $maxbackwardsdate;
+        $periods = [
+            HOURSECS      => get_string('one_hour', 'tool_cloudmetrics'),
```
brendanheywood: Can this be refactored to dedup the same array at:

https://github.com/catalyst/moodle-tool_cloudmetrics/pull/92/files#diff-0b6a2a0fb28e2d23493ed5518e61238fdcae2431af1184bc1b8b1578fda8a007R120

brendanheywood:
> Just checking, in the following example it is set to be hourly, for example between two first rows, should it display between 12:00 and 16:00 (12, 13, 14, 15 , 16) with a 0 value ? Is that what you mean ? I am using pgsql maybe issue is related to mariadb/mysqli

The golden rule is imagine that the metric had been running the whole time and we didn't need to back fill. Ideally the end result should be exactly the same. So yes there would be a metric for every time slice and many of them would be 0

brendanheywood:
Also I am using postgress too

classes/admin_setting_manage_metrics.php
marcghaly: Added

classes/collector/base.php
marcghaly: Fixed.

classes/metric/base.php
marcghaly: Fixed

classes/metric/online_users_metric.php
marcghaly: No it is present in parent class, removed it.

classes/metric/base.php
marcghaly: Would you like to have plain days instead of sliced interval data ? I will document and add changes based on your answer.

classes/metric/online_users_metric.php
marcghaly: Fixed

classes/metric/online_users_metric.php
marcghaly: Fixed

collector/database/classes/collector.php
marcghaly: Fixed.

classes/metric/base.php
marcghaly: Fixed

collector/database/classes/collector.php
marcghaly: Fixed.

collector/database/classes/collector.php
marcghaly: I found it confusing seeing different frequencies in the chart - what happens at the moment : it retrieves all data for the requested time-frame unless the frequency changed, when frequency changes, it deletes all data and replaces it with the new data with proper interval. Also, if I request data for 4 months, it will fill data for those 4 months then if I request data for 12 months, it will retrieve data from the lowest timestamp present in DB to the passed point time requested (with the period) - it will only retrieve the missing part.

collector/database/classes/form/metric_backfill_form.php
marcghaly: I have passed it as part of the 2nd parameter of the moodle form.

collector/database/chart.php
marcghaly: I have split those in two pages chart + backfill

collector/database/chart.php
marcghaly: I have added an exception earlier to the `backfill` page.

marcghaly:
Hi @brendanheywood,

I have made changes according to your comments, If changes are okay I will squash commits.

Cheers,

Marc

classes/metric/base.php
```diff
@@ -125,6 +125,29 @@ public function get_settings_url() {
...
+        }
+        return [$config[0], $config[1], $config[2]];
+    }
+
+    /**
+     * Sets the range for which a backfilled metric has retrieved data.
+     *
+     * @param string $currentconfig Config containing min and max timestamps plus interval.
+     */
+    public function set_range_retrieved(string $currentconfig) {
```
brendanheywood: would it now make sense for set and get to take the same shaped data,ie array of 3 things

classes/metric/base.php
brendanheywood: Not quite sure what you mean by plain days. To me if the metric happens to be once a day then it would be time slices of each day and each day starts at midnight and ends at midnight. But this all depends on the metric frequency, the same applies if it is an hourly metric or every 5 minutes

collector/database/classes/collector.php
brendanheywood: I think deleting data has some business value but I think it should be explicit, similar to the setting for how much data you want to keep around. I don't like the idea of someone trying to add in more data and ending up losing data they already had.

I've split that off into a new issue, if you want you can tackle that after this pr lands:
https://github.com/catalyst/moodle-tool_cloudmetrics/issues/104

For this one I'd just want it to only ever add more data

classes/metric/base.php
marcghaly: Fixed.

collector/database/classes/collector.php
marcghaly: I have set it up so it never deletes data now - but also that if frequency changed it is not overlapping data of another interval - It would get extremely confusing. If frequency changes, timestamp needs to be earlier than lowest one in DB so it completes data, it looks like this. Highlighted the one month period - the other one 1 week.

Screenshot from 2022-06-08 15-54-27

I will sure get on this one once this PR lands.

marcghaly:
Hi @brendanheywood, this is squashed and rebased over latest master.

Regards,

Marc

brendanheywood:
thanks @marcghaly

PR #100, opened 2022-06-01
Monthsecs constant add

marcghaly:
Hi @jaypha,

This is a fix for a constant used in the collector/database/chart.php, 
`Warning: Use of undefined constant MONTHSECS - assumed 'MONTHSECS' (this will throw an Error in a future version of PHP) in /admin/tool/cloudmetrics/collector/database/chart.php on line 79`,
`Warning: A non-numeric value encountered`, it seems this constant is not defined in lib/moodlelib.php.

Regards,

Marc

### DIFF
--- a/collector/database/chart.php
+++ b/collector/database/chart.php
@@ -24,6 +24,7 @@ use tool_cloudmetrics\metric;
  * @copyright 2022, Catalyst IT
  * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
+define('MONTHSECS', WEEKSECS * 4);
 
 require_once(__DIR__.'/../../../../../config.php');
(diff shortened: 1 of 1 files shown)


collector/database/chart.php
```diff
@@ -24,6 +24,7 @@
  * @copyright 2022, Catalyst IT
  * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
+define('MONTHSECS', WEEKSECS * 4);
```
brendanheywood, line 27: I'm not sure we should add this constant as it is misleading, I would prefer to see all uses of this removed instead, I've done this instead:

https://github.com/catalyst/moodle-tool_cloudmetrics/commit/7121d6e845459741e97a6f849c382b63a0a217bb

Issue #106, opened 2022-07-12
Incorrect data returned by cltr::get_enabled_plugins() causes unit test failure in Moodle


katerynadegtyariova:
The current implementation of cltr::get_enabled_plugins() causes unit test failure in Moodle.

Run the following command to repeat:
`vendor/bin/phpunit "core_plugin_manager_testcase" lib/tests/plugin_manager_test.php`

The error output is very long, so it will be truncated here:
```
root@a98befdc0147:/var/www/moodletest# vendor/bin/phpunit "core_plugin_manager_testcase" lib/tests/plugin_manager_test.php
Moodle 3.9.14+ (Build: 20220701), 1b0ca9c4ded8c25141a6a62b728323800ca036c8
Php: 7.4.29, pgsql: 9.6.24, OS: Linux 5.15.0-37-generic x86_64
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

......F..................................                         41 / 41 (100%)

Time: 8.69 seconds, Memory: 93.00 MB

There was 1 failure:

1) core_plugin_manager_testcase::test_get_enabled_plugins
Failed asserting that tool_cloudmetrics\plugininfo\cltr Object &000000000eb286ce00000000759a04e4 (
    'type' => 'cltr'
    'typerootdir' => '/var/www/moodletest/admin/tool/cloudmetrics/collector'
    'name' => 'database'
    'displayname' => 'Cloudmetrics database collector'
   ....
    'availableupdates' => null
) is identical to 'database'.

/var/www/moodletest/lib/tests/plugin_manager_test.php:108
/var/www/moodletest/lib/phpunit/classes/advanced_testcase.php:80
phpvfscomposer:///var/www/moodletest/vendor/phpunit/phpunit/phpunit:60

To re-run:
 vendor/bin/phpunit "core_plugin_manager_testcase" lib/tests/plugin_manager_test.php

FAILURES!
Tests: 41, Assertions: 6504, Failures: 1.
```

katerynadegtyariova:
The issue is fixed by the following Pull Request:

https://github.com/catalyst/moodle-tool_cloudmetrics/pull/109

PR #107, opened 2022-07-12
Issue #106: Incorrect data returned by cltr::get_enabled_plugins() causes unit test failure in Moodle

katerynadegtyariova:
This is PR for Issue #106

The patch fixes the result for cltr::get_enabled_plugins().
The function cltr::get_enabled_plugins() is supposed to return the array of enabled plugins as $pluginname=>$pluginname.
Returning array of plugin objects is incorrect.

The error message and instructions to repeat can be found in the description of Issue #106.

After applying the fix, the test passes successfully:
```
root@a98befdc0147:/var/www/moodletest# vendor/bin/phpunit "core_plugin_manager_testcase" lib/tests/_manager_test.php
Moodle 3.9.14+ (Build: 20220701), 1b0ca9c4ded8c25141a6a62b728323800ca036c8
Php: 7.4.29, pgsql: 9.6.24, OS: Linux 5.15.0-37-generic x86_64
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

.........................................                         41 / 41 (100%)

Time: 5 seconds, Memory: 93.00 MB

OK (41 tests, 6620 assertions)
```

### DIFF
--- a/classes/plugininfo/cltr.php
+++ b/classes/plugininfo/cltr.php
@@ -38,12 +38,18 @@ class cltr extends \core\plugininfo\base {
      */
     public static function get_enabled_plugins() {
         $plugins = \core_plugin_manager::instance()->get_plugins_of_type('cltr');
+
+        if (empty($plugins)) {
+            return array();
+        }
+
+        $enabled = array();
         foreach ($plugins as $name => $plugin) {
-            if (!$plugin->is_enabled()) {
-                unset($plugins[$name]);
+            if ($plugin->is_enabled()) {
+                $enabled[$name] = $name;
             }
         }
-        return $plugins;
+        return $enabled;
     }
 
     /**


classes/plugininfo/cltr.php
```diff
@@ -38,12 +38,18 @@ class cltr extends \core\plugininfo\base {
...
+        if (empty($plugins)) {
+            return array();
+        }
+
+        $enabled = array();
         foreach ($plugins as $name => $plugin) {
-            if (!$plugin->is_enabled()) {
-                unset($plugins[$name]);
+            if ($plugin->is_enabled()) {
+                $enabled[$name] = $name;
```
jaypha, line 49: Hi Kateryna.
The original function returns an array of plugins, but your fix results in returning an array of strings.
After some checking, I realise that the original design is in error. However a lot of the functionality in this plugin relies on this implementation.
Can I ask you to change this to return an array of plugins, so that the plugin does not crash, and I will raise a separate issue for the incompatibility.
Thanks.

katerynadegtyariova:
Thanks Jason. I am closing the PR then.

PR #117, opened 2022-08-16
Issue #114: Turn off metric_expiry by default

### LINKED ISSUE #114: Turn off metric_expiry by default
Make this an upgrade script too so if it is on we turn it off

petersistrom:
Resolves #114


collector/database/settings.php
```diff
@@ -52,7 +52,7 @@
                 'cltr_database/metric_expiry',
                 get_string('metric_expiry', 'cltr_database'),
                 get_string('metric_expiry_desc', 'cltr_database'),
-                30 * DAYSECS
+                0
```
brendanheywood, line 55: There should be a Lang pack tweak somewhere to explain the 0

PR #125, opened 2022-08-18
Issue #74: Show results at different time granularity

### LINKED ISSUE #74: ctlr_db add a setting to show results at a different time granularity
If we have a metric that records every minute, we might want to show a graph that summarized results to every 10 minutes or 1 hour

So add a new setting which could be a second drop down side by side with the metric selector. Its default value will be 'default' which inherit the grain size from the metric and does nothing. But if the grain size if different then pull out the metric and then aggregate it. Instead of showing just the metric itself we will want to show its average over each time slice. and also its minimum and maximum

image

petersistrom:
Resolves #74

### DIFF
--- a/collector/database/chart.php
+++ b/collector/database/chart.php
@@ -23,12 +23,11 @@
  * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 
-use cltr_database\form\metric_backfill_form;
-use cltr_database\collector;
 use core\chart_line;
 use core\chart_series;
-use single_select;
+use tool_cloudmetrics\lib;
 use tool_cloudmetrics\metric;
+use tool_cloudmetrics\metric\manager;
 
 require_once(__DIR__.'/../../../../../config.php');
 require_once($CFG->libdir . '/adminlib.php');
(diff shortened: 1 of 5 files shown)


collector/database/chart.php
```diff
@@ -107,25 +109,55 @@
 
 $periodselect->set_label(get_string('select_graph_period', 'cltr_database'));
 
+$configfrequency = get_config('tool_cloudmetrics', $metricname . '_frequency');
+$selectedfrequency = optional_param('graphfrequency', $configfrequency, PARAM_INT);
```
jaypha: `$selectedfrequency` needs some kind of default value or get_metrics_aggregated will complain.

collector/database/chart.php
```diff
@@ -28,7 +28,9 @@
 use core\chart_line;
 use core\chart_series;
 use single_select;
```
jaypha: `use single_select` is redundant and causes a warning message.

jaypha:
This also needs a version bump because of lang string changes.

collector/database/chart.php
```diff
@@ -107,25 +106,69 @@
 
 $periodselect->set_label(get_string('select_graph_period', 'cltr_database'));
 
+$configfrequency = get_config('tool_cloudmetrics', $metricname . '_frequency');
```
jaypha: You should not use get_config() to obtain or set values associated with metrics. Not all metrics will be defined in the tool_cloudmetrics plugin, so this line line would not work for non-native metrics. Use the metric classes instead.

```suggestion
$configfrequency = $metrics[$metricname]->get_frequency();
```

jaypha:
It's looking good from my perspective. However I noticed that if you change the graph frequency, and then change the graph period, the graph frequency resets. I don't know if this is intentional, but it is a little annoying.

PR #139, opened 2022-08-22
Issue #41: Show multiple metrics on chart

### LINKED ISSUE #41: Database collector should be able to show multiple metrics on the same graph
I'm thinking the chart page has a list down the left of all the metrics with checkboxes for each. You can toggle any combo on and off and bookmark that url.

The metrics should be listed with their schedule, generally you will want to mash together metrics which are mostly the same schedule or within a close range, eg 1 min + 15 minute, but not 1 min with 1 months.

petersistrom:
Resolves #41

### DIFF
--- a/collector/database/chart.php
+++ b/collector/database/chart.php
@@ -41,39 +41,41 @@ $url = new moodle_url('/admin/tool/cloudmetrics/collector/database/chart.php');
 $PAGE->set_context($context);
 $PAGE->set_url($url);
 
-$metricname = optional_param('metric', 'activeusers', PARAM_ALPHANUMEXT);
-
 $defaultperiod = optional_param('graphperiod', -1, PARAM_INT);
+
+$metrics = metric\manager::get_metrics(true);
+$metriclabels = [];
+$checkboxes = [];
+$displayedmetrics = [];
+$notifications = [];
+
+foreach ($metrics as $m) {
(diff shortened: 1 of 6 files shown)


brendanheywood:
Getting a few weird bugs which I assume are all the same root cause:

Duplicate metric names:
https://master.localhost/admin/tool/cloudmetrics/collector/database/chart.php?graphfrequency=512&metric=activeusers
image


Weird query param: double activeusers AND metric=activeusers, it seems we only need one set of params or the other and not both:

chart.php?activeusers=1&activeusers=1&graphfrequency=512&metric=activeusers

collector/database/chart.php
```diff
@@ -113,8 +122,10 @@
     $freqoptions,
     $selectedfrequency
 );
+$checkboxes[] = ['freq' => html_writer::empty_tag('input',
```
jaypha: This appears a little dodgy. It is a single item tacked on to the end of an array it doesn't belong to. It should be added to the template context as its own entity.

collector/database/chart.php
```diff
@@ -113,8 +122,9 @@
     $freqoptions,
     $selectedfrequency
 );
+$checkboxes[] =
```
jaypha: This line needs to be removed.

PR #144, opened 2022-08-25
Issue #140: Performance checks for metrics

### LINKED ISSUE #140: Expose the latest version of each metric via as a performance check
* [ ] this will be the latest cached / stored version of each metric
* [ ] this could be reading from the database collector table OR it could just read / write from config at the cloudmetrics level leaning towards the latter

petersistrom:
Resolves #140

### DIFF
--- a/classes/check/metriccheck.php
+++ b/classes/check/metriccheck.php
@@ -1,0 +1,97 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_cloudmetrics\check;
+
+use core\check\check;
+use core\check\result;
+use tool_cloudmetrics\metric\base;
+use tool_cloudmetrics\metric\manager;
+use tool_cloudmetrics\metric\metric_item;
+
+/**
+ * Performance check for metrics
+ *
+ * @package tool_cloudmetrics
+ * @author Peter Sistrom (petersistrom@catalyst-au.net)
+ * @license http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ * @copyright Catalyst IT
+ */
+class metriccheck extends check {
+
+    /** @var base $metric to be checked*/
+    private $metric;
+
+    /**
+     * Constructor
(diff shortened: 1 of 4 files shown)


classes/check/metriccheck.php
```diff
@@ -0,0 +1,81 @@
...
+    private $metric;
+
+    /**
+     * Constructor
+     *
+     * @param metric_item $metric
+     */
+    public function __construct($metric) {
+        $this->metric = $metric;
+        $this->id = $metric->get_name();
```
brendanheywood, line 45: all checks should have a unique id, at the moment it creates multiple checks with the same id:

tool_cloudmetrics_metriccheck

classes/check/metriccheck.php
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * Return result
+     * @return result
+     */
+    public function get_result() : result {
+        if (!$this->metric->is_enabled()) {
+            return new result(result::INFO, get_string('metric_not_enabled',
+                'tool_cloudmetrics', $this->metric->get_label()));
+        }
+        $value = get_config('tool_cloudmetrics', $this->metric->get_name() . '_last_value');
```
brendanheywood, line 94: nothing is coming out the other side after running cron:

https://master.localhost/report/performance/index.php?detail=tool_cloudmetrics_metriccheck

image

petersistrom:
To test I ran the following code to insert data into the logs 
```
GLOBAL $DB;
$dataobjects = [];
$res = (1661313653 - 1642996853) / 100;
for ($i = 1590465600; $i < 1661313653; $i += $res) {
    $dataobjects[] = [
        'eventname' => '\core\event\user_created',
        'component' => 'core',
        'action' => 'created',
        'target' => 'user',
        'crud' => 'r',
        'edulevel' => 0,
        'contextid' => 1,
        'contextlevel' => 10,
        'contextinstanceid' => 0,
        'userid' => $i,
        'anonymous' => 0,
        'timecreated' => $i
    ];	
}
$DB->insert_records('logstore_standard_log', $dataobjects);
```

Then backfilled Online Users for 2 months and the last value is displayed
image

classes/collector/base.php
```diff
@@ -53,6 +53,9 @@ public function record_metrics(array $metrics, \progress_bar $progress = null) {
                 $count++;
             }
         }
+        if ($count > 0) {
+            set_config($metric->name . '_last_value', $metrics[$count - 1]->value, 'tool_cloudmetrics');
```
brendanheywood: If more than one collector is on then this will run multiple times.

Conversely if no collectors are on then it won't run either, feels like this should be moved up a layer.

classes/check/metriccheck.php
petersistrom: Updated to create a unique id for each check

classes/collector/base.php
petersistrom: Moved setting the config to the task

jaypha:
Screenshot from 2022-08-31 10-52-09

I think it should be more descriptive than just saying '1', like "Last recorded '1' at 12:55", or similar.

brendanheywood:
This is intentional as this is going to get pushed elsewhere for consumption and I want it to be easy to cast to an int or float. So just put that in a code comment

PR #147, opened 2022-08-26
Issue #97: Make chart display the full time period.

### LINKED ISSUE #97: db collectors chart: if you select X window size, force the x axis to show this
ie if there is no data on either end then don't shrink wrap the graph, force the left axis to be 2 weeks ago if you want 2 weeks of data

jaypha:
Resolves #97 

Maximum data points is still limited to 1000.

### DIFF
--- a/collector/database/chart.php
+++ b/collector/database/chart.php
@@ -132,6 +132,7 @@ $aggregatefreqtimes = lib::FREQ_TIMES;
 
 // TODO Handle a month properly currently aggregated over last 30 days.
 $aggregatefreqtimes[4096] = 30 * 24 * 60 * 60;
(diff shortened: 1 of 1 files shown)


petersistrom:
Not sure if this is an issue or not but seems to be working fine if the data is under the 1000 point limit, but the dates are being truncated if we go over the limit
image
Screenshot from 2022-08-26 16-20-32

jaypha:
I did say that it was still limited to displaying 1000 data points. That means that if there are more than 1000 data points over the time period, then it will get truncated.

If this is not desired behaviour, then please specify, and I will update it.

brendanheywood:
I think land this as is, there is more improvements to this in https://github.com/catalyst/moodle-tool_cloudmetrics/issues/148

PR #153, opened 2022-08-30
Issue #148: Change displayed frequency to keep num of displayed points under 1000.

### LINKED ISSUE #148: Improve on the 1000 record limit and automatically reduce the resolution if you zoom out
Lets say you are looking at 1 days worth of data in 5 minute increments, and then zoom out to a week or a month. If you would hit the 1000 record limit then automatically change the increment from 5 minutes up a notch until you are under the 1000 record limit.

I'm on the fence around whether this is an internal safeguard or whether it would change the parameters for you.

jaypha:
Resolves #148

### DIFF
--- a/collector/database/chart.php
+++ b/collector/database/chart.php
@@ -41,8 +41,11 @@ $url = new moodle_url('/admin/tool/cloudmetrics/collector/database/chart.php');
 $PAGE->set_context($context);
 $PAGE->set_url($url);
 
-$defaultperiod = optional_param('graphperiod', -1, PARAM_INT);
+// Graph display timespan, in seconds.
+$graphperiodsec = optional_param('graphperiod', -1, PARAM_INT);
+// Group of metrics to display (if any).
 $metricgroup = optional_param('groupselect', '', PARAM_ALPHANUMEXT);
+
 $metrics = metric\manager::get_metrics(true);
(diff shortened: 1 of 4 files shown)


collector/database/chart.php
```diff
@@ -267,18 +282,23 @@
...
 $context['frequency'] = html_writer::empty_tag('input',
-    array('type' => 'hidden', 'name' => 'graphfrequency', 'value' => $selectedfrequency));
+    array('type' => 'hidden', 'name' => 'graphfrequency', 'value' => $displayfrequency));
 $renderer = $PAGE->get_renderer('tool_cloudmetrics');
 
 echo $OUTPUT->header();
 if ($count == 0) {
     echo $OUTPUT->notification(get_string('norecords', 'cltr_database', $maxrecords), 'notifyproblem');
 } else {
+    echo $OUTPUT->notification(get_string('displaying_records', 'cltr_database', ['count' => $count, 'freq' => $freqoptions[$displayfrequency]]), 'notifysuccess');
```
brendanheywood: We are showing a lot of notifications which are becoming a bit of clutter. Can you swap all of these to 'info' rather than success and can you put them under the graph.

brendanheywood: Rest is great, once this is tweaked merge away

PR #154, opened 2022-08-30
Issue #138: Show default aggregation when chart is overcrowded

### LINKED ISSUE #138: Each metric should decide what aggregation to use when aggregated
If we show an aggregation of a metric, and we don't have room to show min / avg / max only one of them, then each metric should declare which one it should show. In the base this should be max. Very few would override it but there are some metric where using min or avg might make more sense, ie when a lower number is worse rather than higher being worse, and we generally want to show the worse number.

petersistrom:
Resolves #138

### DIFF
--- a/classes/metric/base.php
+++ b/classes/metric/base.php
@@ -25,6 +25,10 @@ namespace tool_cloudmetrics\metric;
  * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 abstract class base {
+
+    /** @var string $aggregate_default Value to display if MIN, MAX and AVG are too close */
+    public $aggregatedefault = 'MAX';
+
     /**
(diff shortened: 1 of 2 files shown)


classes/metric/builtin_user_base.php
```diff
@@ -26,6 +26,9 @@
  */
 abstract class builtin_user_base extends builtin_base {
 
+    /** @var string aggregate_default Value to display if MIN, MAX and AVG are too close */
+    public $aggregatedefault = 'MAX';
```
jaypha: Would this be better in base?

jaypha: Everything else is all good.

petersistrom: Updated

PR #160, opened 2023-07-25
WR408755: Minor changes in .github/workflows/ci.yml for publishing

srdjan-catalyst:


version.php
```diff
@@ -26,7 +26,7 @@
 defined('MOODLE_INTERNAL') || die();
 
 $plugin->version = 2022082500;
-$plugin->release = 2022082500;
+$plugin->release = "2022082500 for Moodle 3.5+";
```
brendanheywood: Do we need the words? Not saying we shouldn't just we should be consistent so if there is a good reason then lets do it across the board

README.md
```diff
@@ -1,7 +1,5 @@
 # moodle-tool_cloudmetrics
 
-[![ci](https://github.com/catalyst/moodle-tool_cloudmetrics/workflows/ci/badge.svg)](https://github.com/catalyst/moodle-tool_cloudmetrics/actions?query=workflow%3Aci)
```
brendanheywood: Why no badge?

srdjan-catalyst: The wiki says that's the old way. I assumed it was added automatically if ci.yml was there. My misinterpretation?

version.php
srdjan-catalyst: We don't, but the release field def says "String describing...". I just thought it was appropriate. Otherwise it's just a copy of version, no added value.

README.md
brendanheywood: We still want the badge, the wiki mentions removing the old travis badge if it is present. I've tweaked the wiki

version.php
brendanheywood: We generally just make it the same purely to make it quick and simple. We have not ever bothered to do a version like 'RC1' or 'Release 1.3.5' because its just more work to maintain for not much value

.github/workflows/ci.yml
```diff
@@ -1,9 +1,13 @@
-# .github/workflows/ci.yml
 name: ci
 
-on: [push, pull_request]
+env:
+  PLUGIN: tool_cloudmetrics
```
brendanheywood: Still not sure why this has changed, either this is right and the docs are wrong and should be fixed or vice versa

https://github.com/catalyst/catalyst-moodle-workflows#add-the-workflow

srdjan-catalyst: https://wiki.wgtn.cat-it.co.nz/wiki/Catalyst_Australia_Pty_Ltd/Elearning/best_practices/publishing-a-plugin#2._Set_.60PLUGIN.60_environment_variable

brendanheywood: Hmm I think the one in github is cannonical and more correct @keevan can you check this? I think I'll delete all the stuff in the wiki

keevan: Yeah we should be using the one here now:
https://github.com/catalyst/catalyst-moodle-workflows

brendanheywood: @srdjan-catalyst I have cleaned up the wiki, we should not need the env: PLUGIN either this can be inferred

brendanheywood: But we DO need the secret as per the github docs

keevan: Hi @srdjan-catalyst ,

You should be able to get the release workflow working by replacing everything here with the default example in the add the workflow section, with some minimal tweaks:
https://github.com/catalyst/catalyst-moodle-workflows#add-the-workflow

- You'll need to uncomment the "secret:" and "moodle_org_token" lines so they are no longer comments)
- You'll need to get rid of the `with:` statement, if there's no options this plugin needs to pass through.

Issue #166, opened 2024-04-10
Implementation of  get_plugins_with_function() is not following Moodle core standard when custom metrics are to be collected from tool_cloudmetrics.


josepico01:
While implementing custom metrics in a local plugin, it was noticed that tool_cloudmetrics is using callback `get_plugins_with_function('_metrics', lib.php) ` to find those custom metrics in a third party plugin where a function with the suffix `metrics` is declared in the lib.php . However, no custom metrics are found because of the `_` character added as prefix in the function parameter. 

Looking at other callbacks from core, none of those are with a `_` character as part of the function's name, so this character should be removed for the tool to be able to collect those custom metrics.

PR #167, opened 2024-04-10
Issue #166: Fix callback to pick up custom metrics from third party plugins

josepico01:
This fix will allow the `tool_cloudmetrics` to recognised custom metrics from third party plugins where a function with the suffix metrics is declared inside the` lib.php` file.


brendanheywood:
This should ideally also get coverted to the new hook api

josepico01:
Thanks for the speedy response on this one @brendanheywood

Issue #168, opened 2024-04-16
When implementing custom metrics with camelcase names, the graphs API does not showing data for these due to mismatched names. 


josepico01:
While  implementing custom metrics, it was found that camelcase metric's names are not supported by the graph API in `tool_cloudmetrics` and the reason is that the metric's name is passed as an identifier in queries (not double-quoted), therefore it is transformed into a lowercase column and will not fetch data for the corresponding metric.

As a result, a warning error is thrown when debugging is on Dev mode.

PR #170, opened 2024-04-17
Issue #168: Fix metric identifiers in chart API to remain unchanged

josepico01:
- This change will allow the charts API to support camelcase metric names.

### DIFF
--- a/collector/database/classes/collector.php
+++ b/collector/database/classes/collector.php
@@ -125,15 +125,15 @@ class collector extends base {
         }
 
         if ($aggregate == DAYSECS) {
-            $incrementstart = "date as increment_start";
+            $incrementstart = "date AS increment_start";
         } else {
-            $incrementstart = "FLOOR(time/$aggregate) * $aggregate as increment_start";
+            $incrementstart = "FLOOR(time/$aggregate) * $aggregate AS increment_start";
         }
         list ($clause, $params) = $DB->get_in_or_equal($metricnames);
         if (count($metricnames) == 1) {
-            $sql = "SELECT AVG(" . $DB->sql_cast_char2int('value', true) . ") as \"$metricnames[0]\",
-                MIN(" . $DB->sql_cast_char2int('value', true) . ") as min,
-                MAX(" . $DB->sql_cast_char2int('value', true) . ") as max,
+            $sql = "SELECT AVG(" . $DB->sql_cast_char2int('value', true) . ") AS \"$metricnames[0]\",
+                MIN(" . $DB->sql_cast_char2int('value', true) . ") AS min,
+                MAX(" . $DB->sql_cast_char2int('value', true) . ") AS max,
                 $incrementstart
                 FROM {cltr_database_metrics}
                 WHERE name $clause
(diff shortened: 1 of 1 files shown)


collector/database/classes/collector.php
```diff
@@ -143,7 +143,7 @@ public function get_metrics_aggregated($metricnames = null, $since = false, int
         } else {
             $metricselect = '';
             foreach ($params as $param) {
-                $metricselect .= "AVG(CASE name WHEN '$param' THEN" . $DB->sql_cast_char2int('value', true) . "END) $param,";
+                $metricselect .= "AVG(CASE name WHEN '$param' THEN" . $DB->sql_cast_char2int('value', true) . "END) \"$param\",";
```
cameron1729: This SQL looks odd to me. Does it actually work? It looks like it's missing an `AS`.

josepico01: It is working without it but following your suggestion, I added it to the SQL. Thanks.

cameron1729: I think it would fail when using Oracle. **Edit**: NVM, also optional in Oracle. But in any case, I think the SQL is easier to understand with it (especially since the `$param` variable makes it unclear what's happening). Plus the other query above has it too.

While we're playing with it, it's really odd that it uses `get_in_or_equal` because the `$clause` part is never used. Would we be able to change it to do something like `foreach $metricnames as $metricname {` and get rid of the `get_in_or_equal`. It looks like it's a copy+paste error.

collector/database/classes/collector.php
```diff
@@ -143,7 +143,7 @@ public function get_metrics_aggregated($metricnames = null, $since = false, int
         } else {
             $metricselect = '';
             foreach ($params as $param) {
-                $metricselect .= "AVG(CASE name WHEN '$param' THEN" . $DB->sql_cast_char2int('value', true) . "END) $param,";
+                $metricselect .= "AVG(CASE name WHEN '$param' THEN" . $DB->sql_cast_char2int('value', true) . "END) as \"$param\",";
```
cameron1729: [Moodle SQL coding style](https://moodledev.io/general/development/policies/codingstyle/sql) specifies that all SQL keywords should be uppercase. Can you update it to "AS" please :) (might as well fix the [other query](https://github.com/catalyst/moodle-tool_cloudmetrics/pull/170/files#diff-3dc9a582fde560b17b20084cacf6a04f1b2955786117b996feaa4af2ef8cae92R134) while we're here too).

collector/database/classes/collector.php
josepico01: Agree with the SQL changes. 
Regarding the `get_in_or_equal()` the $clause is actually used  [here](https://github.com/catalyst/moodle-tool_cloudmetrics/blob/MOODLE_35_STABLE/collector/database/classes/collector.php#L92). Would you still recommend removing it?

collector/database/classes/collector.php
josepico01: Thanks for pointing this out. I have updated these keywords to be uppercase.

collector/database/classes/collector.php
cameron1729: Oops, I'm blind :) Let's leave it then.

Issue #169, opened 2024-04-16
Unit test fails in tool_cloudmetrics when implementing custom metrics


josepico01:
`tool_cloudmetrics\tool_cloudmetrics_collect_metrics_test::test_execute` failed.

The test file is basically measuring time, frequency for the list of metrics received when executing the schedule task `\tool_cloudmetrics\task\collect_metrics_task` and passing this to a mock class.

Since the task is implementing the `manager::get_metrics()` function, not only the inbuilt metrics are fetched but also the custom metrics. But, the provider function is hard coding the expected metrics to be only from the inbuilt list.

PR #171, opened 2024-04-17
Issue #169: Update unit test  to disable custom metrics.

josepico01:
- Since all metrics are enabled by default and data provider was supplied with builtin metrics only, we need to disable those custom metrics before the schedule task is executed.

### DIFF
--- a/tests/tool_cloudmetrics_collect_metrics_test.php
+++ b/tests/tool_cloudmetrics_collect_metrics_test.php
@@ -138,8 +138,10 @@ class tool_cloudmetrics_collect_metrics_test extends \advanced_testcase {
 
     /**
      * This method gets all metrics (except those in sent in the data provider) and sets them into a disable state.
+     * @param array $expected The expected metrics.
+     *
      */
-    private function disable_metrics(array $expected): void {
+    private function disable_metrics(array $expected) {
         $metrics = metric\manager::get_metrics(true);
 
         foreach ($metrics as $metric) {
--- a/tests/tool_cloudmetrics_collect_metrics_test.php
+++ b/tests/tool_cloudmetrics_collect_metrics_test.php
@@ -126,11 +126,30 @@ class tool_cloudmetrics_collect_metrics_test extends \advanced_testcase {
         $mock->expects($this->once())
             ->method('receive')
             ->with($expected);
+
+        $this->disable_custom_metrics();
         $task = new helper_collect_metrics_task($mock);
         $task->set_time($time);
         ob_start();
         $task->execute();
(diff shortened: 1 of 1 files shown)


cameron1729:
Thanks @josepico01 

With this patch, what would happen if a new "built in" metric was added and the `execute_provider` remained unchanged? I think we'd have the same problem right?

I think a better approach may be to disable all metrics at the beginning of `test_execute` - and this may have been the original intention because there's [a line](https://github.com/catalyst/moodle-tool_cloudmetrics/pull/171/files#diff-dcb6fc97f9bad84acfa470ffb92463074eb0249768fa8091e31fd9ef84476231R121) in the loop that explicitly enables the metric. By doing it this way, all metrics are gauranteed to be disabled, except the ones provided by `execute_provider`.

What do you think?

tests/tool_cloudmetrics_collect_metrics_test.php
```diff
@@ -126,13 +126,32 @@ public function test_execute(string $timestr, array $meta, array $expected) {
...
         $task->set_time($time);
         ob_start();
         $task->execute();
         ob_end_clean();
     }
 
+    /**
+     * This method gets all metrics available and disable those out of the builtin metrics.
+     *
+     * @return void
```
cameron1729: Can remove this return annotation :)

josepico01: Thanks for the feedback. The metrics are required to be measured in sync with each other reason why the test will fail again if I disable all of them. For example, the last set of data provider sends the `new_user_metric`  with FREQ_5MIN to be measured in the period of time 1 day, and those metrics with default FREQ_5MIN and up to FREQ_DAY will also be measured at this time (which is the case for all builtin metrics). And so on. 

By default, our custom metrics are FREQ_5MIN meaning they should be process by the task as well. So, I maybe need to find another approach around.

cameron1729: We could try something like the following:

```
foreach (metric\manager::get_metrics(true) as $metric) {
    if (!in_array($metric->get_name(), $expected)) {
        $metric->set_enabled(false);
    }
}
```

So now, all metrics except the ones that the testcase cares about are disabled. We could also remove the call to `$metric->set_enabled(true)` in the `foreach ($meta as $metric => $data)` loop.

josepico01: Thanks for the suggestion @cameron1729 I have added the fixes and also updated some other code sections to comply with phpcs rules. 
Please review it when available.

PR #173, opened 2024-06-17
Issue #172: Update metric name param type to fix enable/disable functionality for custom metrics.

### LINKED ISSUE #172: Enable / Disable function (eye icon) of the main setting page does not work for custom metrics
The required param for the name of the metric has been set as a plugin type (`PARAM_PLUGIN` ) when validating the param type around  [Line 32](https://github.com/catalyst/moodle-tool_cloudmetrics/blob/MOODLE_35_STABLE/metrics.php#L32) . This ends up throwing an error when a custom metric is disabled/enabled from this page.

Ideally, the plugin should handle metric names in different formats from this settings page as well.

josepico01:
This change should fix the enable/disable functionality for custom metrics.

Closes #172

### DIFF
--- a/metrics.php
+++ b/metrics.php
@@ -29,7 +29,7 @@ require_once('../../../config.php');
 require_once($CFG->libdir.'/adminlib.php');
 
 $action = required_param('action', PARAM_ALPHANUMEXT);
-$name   = required_param('name', PARAM_PLUGIN);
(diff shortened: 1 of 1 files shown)


brendanheywood:
thanks @josepico01 that makes sense to me

PR #189, opened 2025-07-23
Have an option for a metric to declare that it can populate retrospectively to backfill data #67

### LINKED ISSUE #67: Have an option for a metric to declare that it can populate retrospectively to backfill data

duyhuynhdev:
Have an option for a metric to declare that it can populate retrospectively to backfill data #67
Closes #67

### DIFF
--- a/classes/task/autobackfill_metrics_task.php
+++ b/classes/task/autobackfill_metrics_task.php
@@ -1,0 +1,80 @@
+<?php
+// This file is part of Moodle - https://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <https://www.gnu.org/licenses/>.
+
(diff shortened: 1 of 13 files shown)


classes/task/autobackfill_metrics_task.php
```diff
@@ -0,0 +1,75 @@
...
+                'Generating metrics for %s from %s to %s',
+                $metrictype->get_name(),
+                userdate($nowts, '%e %b %Y, %H:%M'),
+                userdate(($nowts - $collectingperiod), '%e %b %Y, %H:%M')
+            ));
+            $metricitems = $metrictype->generate_metric_items($collectingperiod, $nowts);
+            $items = array_merge($items, $metricitems);
+            mtrace(sprintf('Generated %s %s metrics', count($metricitems),  $metrictype->get_name()));
+        }
+        $this->backfill_metrics($items);
```
brendanheywood: I think the slow bit is the inserting of the metrics on this line, and the logs are a little misleading as it says:

Generating metrics for yearlyactiveusers from 23 July 2025, 16:10 to 23 July 2024, 16:10
Generated 366 yearlyactiveusers metrics

But it hasn't actually saved them until later, I think it should insert them inside the loop and only emit the log when it has finished

Issue #199, opened 2025-08-19
Error when Initialising Moodle PHPUnit test environment...


x-iy:
```
The 'onlineusers' metric does not support auto backfilling
The 'activeusers' metric does not support backfilling data
The 'newusers' metric does not support backfilling data
Generating metrics for dailyusers from 19 Aug 2025, 13:04 to 19 Aug 2024, 13:04
!!! Coding error detected, it must be fixed by a programmer: A lock was created but not released at:
[dirroot]/lib/classes/task/manager.php on line 812

 Code should look like:

 $factory = \core\lock\lock_config::get_lock_factory('type');
 $lock = $factory-&gt;get_lock(cron_adhoc_1);
 $lock-&gt;release();  // Locks must ALWAYS be released like this.

 !!!
!! 
Error code: codingerror !!
!! Stack trace: * line 115 of /lib/classes/lock/lock.php: core\exception\coding_exception thrown
* line 156 of /admin/tool/phpunit/cli/util.php: call to core\lock\lock-&gt;__destruct()
 !!
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Nothing to install, update or remove
Generating autoload files
53 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
Default exception handler: Coding error detected, it must be fixed by a programmer: A lock was created but not released at:
[dirroot]/lib/classes/task/manager.php on line 812

 Code should look like:

 $factory = \core\lock\lock_config::get_lock_factory('type');
 $lock = $factory-&gt;get_lock(cron_adhoc_1);
 $lock-&gt;release();  // Locks must ALWAYS be released like this.
```

PR #200, opened 2025-08-19
Fix for #199

x-iy:
Fix for #199

- include check for phpunit on install task run

### DIFF
--- a/classes/metric/online_users_metric.php
+++ b/classes/metric/online_users_metric.php
@@ -106,18 +106,18 @@ class online_users_metric extends builtin_user_base {
         ];
 
         $interval = $secondsinterval[$frequency];
-        $sql = 'WITH user_data AS (
+        $sql = "WITH user_data AS (
                   SELECT floor(timecreated / :interval ) * :intervaldup AS time, userid
                     FROM {logstore_standard_log}
                    WHERE timecreated >= :starttime
                      AND timecreated <= :finishtime
-                     AND action = "loggedin"
+                     AND action = 'loggedin'
                 )
 
                 SELECT user_data.time as time, COUNT(DISTINCT(user_data.userid)) as value
                   FROM user_data
               GROUP BY user_data.time
-              ORDER BY user_data.time ASC';
+              ORDER BY user_data.time ASC";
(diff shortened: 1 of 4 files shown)


classes/metric/online_users_metric.php
```diff
@@ -106,18 +106,18 @@ public function generate_metric_items($backwardperiod, $finishtime = null, ?\pro
         ];
 
         $interval = $secondsinterval[$frequency];
-        $sql = 'WITH user_data AS (
+        $sql = "WITH user_data AS (
```
dmitriim, line 109: Would be nice to add unit tests to cover related method.

x-iy: Have added unit test

dmitriim:
@x-iy  Can you please bump version so it's pushed to Moodle plugin directory automatically. Thanks!

tests/tool_cloudmetrics_online_users_metric_test.php
```diff
@@ -0,0 +1,116 @@
...
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+final class tool_cloudmetrics_online_users_metric_test extends \advanced_testcase {
+
+    /**
+     * Set up before each test
+     */
+    protected function setUp(): void {
+        global $DB;
+        $DB->delete_records(lib::TABLE);
```
dmitriim: why do we delete records, then call parent and then resetAfterTest. Shouldn't it be like 

1. parent::setUp();
2. $this->resetAfterTest();
3. $DB->delete_records(lib::TABLE);

?

dmitriim: But IMHO this  should  go to test_generate_metric_items rather than setUp

x-iy: Looking at this again, we don't need the collector parts as we are just testing generation of online user metric items. The initial code was influenced by this test:
https://github.com/catalyst/moodle-tool_cloudmetrics/blob/MOODLE_404_STABLE/collector/database/tests/cltr_database_test.php#L162

PR #232, opened 2026-07-21
223 backfill bug

### LINKED ISSUE #213: Arguments to backfill metric functions are weak
The backfill functions for metrics take a $backwardperiod and a $finishtime as arguments. The backwardperiod is relative to the current time (as returned by time()), rather than the finishtime. This is confusing.

It can also cause problems. The time() function is used multiple times throughout the action of getting backfill data. If the clock ticks in the meantime, it may cause an error. Unit testing will fail occasionlly.

A better, more stable alternative would be to use start and finish times as arguments and have higher up functions figure out what they should be.

### LINKED ISSUE #223: Backfill bug when time periods are missing
If we have missing time periods, and you run a back fill, then it should fill in any gaps

ie it checks that every date / time slot has a value 

Image

jaypha:
Resolves #223 
Resolves #213


classes/metric/online_users_metric.php
```diff
@@ -59,20 +59,16 @@ public function is_backfillable(): bool {
     /**
      * Generates a number of metric items for a period of time.
      *
-     * @param int $backwardperiod Time period to draw data from (relative to now).
-     * @param int|null $finishtime The end time to draw data from. Defaults to now.
+     * @param int $start_time The start time to draw data from.
+     * @param int|null $finish_time The end time to draw data from. Defaults to now.
```
bwalkerl, line 63: Underscores in names

classes/metric/test_metric.php
```diff
@@ -158,21 +158,19 @@ public function generate_metric_item($starttime, $finishtime): metric_item {
     /**
      * Retrieve multiple metrics.
      *
-     * @param int $backwardperiod
-     * @param ?int $finishtime
+     * @param int $start_time
+     * @param ?int $finish_time
```
bwalkerl, line 162: Underscores

classes/task/autobackfill_metrics_task.php
```diff
@@ -43,160 +45,181 @@ public function get_name() {
     }
 
     /**
-     * Will backfill the metrics using collectors.
+     *  Execute task
```
bwalkerl, line 48: Extra space

classes/task/autobackfill_metrics_task.php
```diff
@@ -43,160 +45,181 @@ public function get_name() {
...
-                    $collector->record_metrics($metricitems, $progress);
-                    if ($collector->is_readable()) {
-                        $collector->set_last_backfilled_frequency($metricclass->get_name(), $metricclass->get_frequency());
-                    }
+            // Get all enabled backfillable metrics.
+            $metrics = [];
+            $allmetrics = metric\manager::get_metrics(true);
+            foreach ($allmetrics as $name => $m) {
+                if ($m->is_backfillable()) {
+                    $metrics[$name] = $m;
```
bwalkerl, line 83: Should this also check `is_autobackfill()` like before?

tests/helper_collect_metrics_task.php
```diff
@@ -53,7 +53,10 @@ public function __construct(mock_receiver $mock) {
     public function send_metrics(array $items) {
         $names = [];
         foreach ($items as $item) {
-            $names[] = $item->name;
+            // Do not include test metrics.
+            if ($item->name != 'foobar') {
```
bwalkerl, line 57: Can we make this not rely on a hardcoded name?

classes/task/autobackfill_metrics_task.php
```diff
@@ -43,160 +45,181 @@ public function get_name() {
...
+     * Transfers metric data from a gap to the collector.
+     *
+     * @param metric\base $metric
+     * @param collector $collector
+     * @param int $low
+     * @param int $high
+     * @return void
+     */
+    public function transfer(metric\base $metric, collector $collector, int $low, int $high) {
+        mtrace("  Filling gap from $low to $high");
```
bwalkerl, line 208: Looks like this space is used for formatting, if so, is there a better way to format it?

classes/task/autobackfill_metrics_task.php
```diff
@@ -43,160 +45,181 @@ public function get_name() {
...
+
+        // Records are in reverse chronological order, matching our descending iteration,
+        // so we can walk both in lockstep.
+        $backrecords->rewind();
+
+        $first = $last = null;
+        for (
+            $currenttime = $endtime;
+            $currenttime >= $starttime;
+            $currenttime = lib::get_previous_time($currenttime, $freq)
```
bwalkerl, line 168: The name `$currenttime` can be confusing since it's not the current time. Can we use something else?

collector/database/classes/collector.php
```diff
@@ -18,7 +18,8 @@
 
 use tool_cloudmetrics\metric\metric_item;
 use tool_cloudmetrics\collector\readable_base;
-use tool_cloudmetrics\metric;
+use tool_cloudmetrics\metric\base as metric;
```
bwalkerl, line 21: I don't think this is used?

classes/task/autobackfill_metrics_task.php
bwalkerl, line 39: No longer used, should be able to be removed

classes/task/autobackfill_metrics_task.php
```diff
@@ -43,160 +45,181 @@ public function get_name() {
...
+        $progressbar->create();
+        if ($metric->is_backfill_incremental()) {
+            // Incremental metrics are stored one at a time.
+            $roughtotal = ($high - $low) / lib::get_period($metric->get_frequency()) + 1;
+            $count = 0;
+            foreach ($items as $item) {
+                $collector->record_metric($item);
+                $progressbar->update(++$count, $roughtotal, '');
+            }
+            $progressbar->update($roughtotal, $roughtotal, '');
```
bwalkerl, line 220: For `yearly_active_users_metric` we aren't getting descriptive updates of the current status, which is important since these can be quite long. This would be because progressbar isn't being passed, was there a reason for this?

bwalkerl (commented):
Thanks @jaypha this worked in my testing and the code is looking much cleaner than before. 

Just a few comments

classes/task/autobackfill_metrics_task.php
jaypha: How about reftime (reference time)?

bwalkerl: I'm fine with anything that doesn't have double meanings, so should be OK.

classes/task/autobackfill_metrics_task.php
jaypha: done

classes/task/autobackfill_metrics_task.php
jaypha: done

classes/task/autobackfill_metrics_task.php
jaypha: done

classes/task/autobackfill_metrics_task.php
jaypha: just using $time. Done

classes/task/autobackfill_metrics_task.php
jaypha: Removed the space.

classes/task/autobackfill_metrics_task.php
jaypha: Not quite following you here. If you are referring to the $msg parameter, I have added some in.

collector/database/classes/collector.php
jaypha: done

classes/task/autobackfill_metrics_task.php
bwalkerl: It's probably the msg param. For example, onlineusers has:

`Saving metric items for onlineusers.13 Aug 2025, 04:15`

While yearlyactiveusers has nothing (and no longer has individual mtraces).  

I thought this was using the passed $progress bar, but on further inspection it looks to be from record_metrics.

tests/helper_collect_metrics_task.php
jaypha: I improved it a little to only include built in metrics

classes/task/autobackfill_metrics_task.php
bwalkerl: I like the changes, but unsure on the 'isupdate' terminology. Should that use 'upgrade' instead? i.e. 'isupgrade' or 'fromupgrade'?

